PR #5, opened 2014-12-09
Datatypes

touchdown:
@marmbrus please review: I believe it is missing long, enum and union of type Seq(t1) where t1 is not null.


vlyubin:
This is cool! We recently merged a pull request that supports almost every avro data type (with the exception of complex unions), so this functionality is now included in spark-avro.

PR #12, opened 2015-01-15
Adding support for the new datatypes

vlyubin:
Adds support for new datatypes. I added a test avro file and wrote a test to ensure stuff works as expected. If you think it's a good idea, I can also add files that I used to generate test.avro, in case someone needs to update it later.

cc @aarondav @marmbrus


README.md
```diff
@@ -94,5 +119,4 @@ JavaSchemaRDD episodes = AvroUtils.avroFile(sqlContext, "episodes.avro");
 ```
 
 ## Building From Source
-This library is built with [SBT](http://www.scala-sbt.org/0.13/docs/Command-Line-Reference.html), which is automatically downloaded by the included shell script.  To build a JAR file simply run `sbt/sbt package` from the project root.
-
+This library is built with [SBT](http://www.scala-sbt.org/0.13/docs/Command-Line-Reference.html), which is automatically downloaded by the included shell script.  To build a JAR file simply run `sbt/sbt package` from the project root. To run the tests, you should run `sbt/sbt test`
```
aarondav: period

src/main/scala/com/databricks/spark/avro/AvroRelation.scala
```diff
@@ -40,8 +46,31 @@ case class AvroRelation(location: String)(@transient val sqlContext: SQLContext)
...
+      case m: Map[Any, Any] => m.map(x => (x._1.toString, convertToSparkSQL(x._2)))
+      case a: GenericData.Array[Any] => a.map(x => convertToSparkSQL(x))
+      case f: Fixed => f.bytes.clone
+      case e: EnumSymbol => e.toString
+      case r: Record => Row.fromSeq((0 until r.getSchema.getFields.size).map { i =>
+        convertToSparkSQL(r.get(i))
+      })
+      case h: ByteBuffer => {
+        val ar = new Array[Byte](h.remaining)
+        var idx = 0
```
aarondav: I think you should use h.get(ar) instead

src/main/scala/com/databricks/spark/avro/AvroRelation.scala
```diff
@@ -40,8 +46,31 @@ case class AvroRelation(location: String)(@transient val sqlContext: SQLContext)
     convertedSchema
   }
 
+  def convertToSparkSQL(obj: Any): Any = {
+    obj match {
+      case u: Utf8 => u.toString
+      case m: Map[Any, Any] => m.map(x => (x._1.toString, convertToSparkSQL(x._2)))
```
aarondav: Maybe add a comment to this method on the special-casing that's going on. Namely,
- Maps always have String keys
- Byte arrays and ByteBuffers are reused by Avro, so we must copy them out.

src/main/scala/com/databricks/spark/avro/AvroRelation.scala
```diff
@@ -40,8 +46,31 @@ case class AvroRelation(location: String)(@transient val sqlContext: SQLContext)
...
+      case a: GenericData.Array[Any] => a.map(x => convertToSparkSQL(x))
```
aarondav: .map() is still doing the Scala lambda allocation in a per-element basis, which is something we should avoid. We should similarly avoid it within buildScan, inside the baseRdd lambda. 

We could leave this optimization for future work, though, as there are a couple other places we can improve if we go down that path. For instance, rather than using Row.fromSeq() for every row, we can construct a SpecificMutableRow which can be reused across rows. Additionally, we could use the schema to generate an array of functions which convert the expected Avro type to the Spark SQL type, rather than using runtime type checking.

These optimizations would likely be made in tandem with some sort of benchmark to ensure that they are in fact significant improvements, and to ensure future patches do not regress performance.

src/main/scala/com/databricks/spark/avro/AvroRelation.scala
```diff
@@ -106,15 +135,28 @@ case class AvroRelation(location: String)(@transient val sqlContext: SQLContext)
...
-        case other =>
-          sys.error(s"Union types with anything other than null not supported: $other")
+      case MAP =>
+        val schemaType = toSqlType(avroSchema.getValueType)
+        SchemaType(
+          MapType(StringType, schemaType.dataType, valueContainsNull = schemaType.nullable),
+          nullable = false)
+
+      case UNION => {
+        if (avroSchema.getTypes.map(x => x.getType).contains(NULL)) {
```
aarondav: nit: should use same variable name here and in the filter (or just `_`). Also, even more nitty, but for consistency I might also use `.exists(_.getType == NULL)` so the logic looks more analogous.

src/main/scala/com/databricks/spark/avro/AvroRelation.scala
```diff
@@ -106,15 +135,28 @@ case class AvroRelation(location: String)(@transient val sqlContext: SQLContext)
...
+          // In case of a union with null, eliminate it and make a recursive call
+          toSqlType(Schema.createUnion(
+            avroSchema.getTypes.filterNot(elm => elm.getType == NULL))).copy(nullable = true)
+        } else avroSchema.getTypes.toSeq match {
+          case Seq(t1, t2) if Set(t2.getType, t1.getType) == Set(INT, LONG) =>
```
aarondav: Type has a getType method? That's pretty weird. Why not do toSet() and match on that, though?

src/main/scala/com/databricks/spark/avro/AvroRelation.scala
vlyubin: Ok, I'll get rid of some of them.

src/main/scala/com/databricks/spark/avro/AvroRelation.scala
vlyubin: Ok, I'll add a comment.

src/main/scala/com/databricks/spark/avro/AvroRelation.scala
vlyubin: avroSchema.getTypes doesn't return types, but rather some internal notion of types (e.g. org.apache.avro.Schema.IntType / LongType, etc.), but apparently they are protected or something, so I can't directly use them. However, if you call getType on them, you actually get Schema.Type.SOMETHING that you can use.

Oh, true, I didn't notice that. It appears though that you can't do match with sets in scala (unless you do some weird hacks). Or I am wrong :) Either way, I made this slightly shorter.

vlyubin:
Performance optimizations and benchmarking will be done in the next iteration.

README.md
```diff
@@ -50,6 +50,31 @@ These examples use an avro file available for download [here](https://github.com
...
+enum -> StringType
+array -> ArrayType
+map -> MapType
+fixed -> BinaryType
+```
+
+As for unions, we only support three kinds of unions:
+1) union(int, long)
+2) union(float, double)
+3) union(something, null), where something is one of the avro types mentioned above, including two types of unions.
```
pwendell: Just an observation, but if we want to support generic union types down the road we can pull it into a struct type:

```
a: union(A, B)
a: struct {
  left: A
  right: B
}
```

Not sure how many people use unions.

aarondav: That would be left, right, righter, righterer, etc., I think, but it should work.

src/main/scala/com/databricks/spark/avro/AvroRelation.scala
```diff
@@ -40,8 +46,40 @@ case class AvroRelation(location: String)(@transient val sqlContext: SQLContext)
     convertedSchema
   }
 
+  // There are several properties of Avro that we have to account for in this method:
+  // 1) Avro uses Utf8 strings, so we want to convert them to java.lang.String for SparkSQL,
+  //    including making the appropriate recursive calls.
+  // 2) Avro map keys are always strings, so we don't need to recurse on them when processing maps.
+  // 3) Byte arrays and ByteBuffers are reused by Avro, so we must copy them out.
+  def convertToSparkSQL(obj: Any): Any = {
```
aarondav: make this private and convert the above to JavaDoc (`/** */`)

src/main/scala/com/databricks/spark/avro/AvroRelation.scala
```diff
@@ -40,8 +46,40 @@ case class AvroRelation(location: String)(@transient val sqlContext: SQLContext)
...
+    obj match {
+      case u: Utf8 => u.toString
```
aarondav: stylistic comment: typically when some elements of the case require a new line, we convert all of them to a new line to improve readability.

Additionally, braces `{ }` are not required for this situation.

src/main/scala/com/databricks/spark/avro/AvroRelation.scala
```diff
@@ -40,8 +46,40 @@ case class AvroRelation(location: String)(@transient val sqlContext: SQLContext)
...
+      case e: EnumSymbol => e.toString
+      case r: Record => Row.fromSeq((0 until r.getSchema.getFields.size).map { i =>
+        convertToSparkSQL(r.get(i))
+      })
+      case h: ByteBuffer => {
+        val ar = new Array[Byte](h.remaining)
+        h.get(ar)
+        ar
+      }
+      case other => other
```
aarondav: I think we may want to avoid accidentally keeping around types we don't expect; maybe change this to

`case primitive: AnyVal => primitive`

and add a `case other` which throws an exception.

src/main/scala/com/databricks/spark/avro/AvroRelation.scala
```diff
@@ -40,8 +46,40 @@ case class AvroRelation(location: String)(@transient val sqlContext: SQLContext)
...
+  // 3) Byte arrays and ByteBuffers are reused by Avro, so we must copy them out.
+  def convertToSparkSQL(obj: Any): Any = {
+    obj match {
+      case u: Utf8 => u.toString
+      case m: Map[Any, Any] => m.map(x => (x._1.toString, convertToSparkSQL(x._2)))
+      case a: GenericData.Array[Any] => {
+        val ar = new Array[Any](a.size)
+        var idx = 0
+        while (idx < a.size) {
+          ar(idx) = convertToSparkSQL(a(0))
```
aarondav: should this be a(idx)? Also, consider more descriptive names (something like a -> avroArray; ar -> javaArray)

src/test/scala/com/databricks/spark/avro/AvroSuite.scala
```diff
@@ -43,4 +45,70 @@ class AvroSuite extends FunSuite {
...
+    val all = TestSQLContext
+      .avroFile(testFile)
+      .collect()
+    assert(all.size == 3)
+
+    val str = TestSQLContext
+      .avroFile(testFile)
+      .select('string)
+      .collect()
+    assert(str(1)(0) == "Terran is IMBA!")
```
aarondav: I knew it!

src/main/scala/com/databricks/spark/avro/AvroRelation.scala
vlyubin: Apparently you cannot use AnyVal in match statements. I can make a set with primitive types and check if other's type belongs there though (or there is a shorter but uglier solution).

However, it seems to me that this is unnecessary. This function is only called for whatever we parsed from avro file. If there was invalid type in the avro file, we would get an appropriate exception when reading it (I checked). As a result, this function will only get legal avro types. We are handling all of them, so no problem should occur. If there was was a union(A,B,C), the obj parameter that this function gets would be one of A,B,C, rather some class that represents unions. If the union wasn't supported by us, we would throw an exception during schema parsing, which happens before.

So I think we should't be adding this. If you still think otherwise, I'll add it.

src/main/scala/com/databricks/spark/avro/AvroRelation.scala
vlyubin: Ok, will do.

src/main/scala/com/databricks/spark/avro/AvroRelation.scala
vlyubin: Oh, good catch! Thanks!

src/main/scala/com/databricks/spark/avro/AvroRelation.scala
```diff
@@ -15,21 +15,27 @@
  */
 package com.databricks.spark.avro
 
+import java.util.Map
```
aarondav: nit: import order should be lexicographic, here and below

aarondav:
LGTM, @marmbrus would you mind taking a look?

src/main/scala/com/databricks/spark/avro/AvroRelation.scala
```diff
@@ -15,21 +15,27 @@
...
+import org.apache.spark.sql.sources.TableScan
 
 import scala.collection.JavaConversions._
 
 case class AvroRelation(location: String)(@transient val sqlContext: SQLContext) extends TableScan {
 
-  val schema = {
+  override val schema = {
     val fileReader = newReader()
     val convertedSchema = toSqlType(fileReader.getSchema).dataType match {
```
marmbrus: We should probably note in the documentation that the code currently assumes the schema is the same for all of the avro files in a directory.

marmbrus:
LGTM

src/main/scala/com/databricks/spark/avro/AvroRelation.scala
```diff
@@ -66,7 +107,7 @@ case class AvroRelation(location: String)(@transient val sqlContext: SQLContext)
     val fs = FileSystem.get(path.toUri, sqlContext.sparkContext.hadoopConfiguration)
 
     val status = fs.getFileStatus(path)
-    val singleFile = if (status.isDir) {
+    val singleFile = if (status.isDirectory) {
```
aarondav: isDirectory is only available in hadoop 2+. Does this project have a hard dependency on Hadoop 2+? If not, we should use isDir, despite being deprecated, to avoid breaking users of older Hadoop versions.

aarondav:
Merged, thanks!

PR #13, opened 2015-01-16
Adding map type and supporting recursive data type casting

timyitong:
Right now, spark-avro does not support Avro Map type while spark sql does contain MapType. This pull request added this function.

Another improvement is that I made the data casting mechanism recursively resolve nested structures, such as an array of records, a map of multiple arrays and etc. Of course, this does not mean those nested fields can be queried over Spark SQL. But at least it could make most Avro data loadable as SchemaRDD. And my understanding is that SchemaRDD should be more universal than Spark SQL. Being able to load nested Avro data structures could help Avro make serialization/deserialization easier. We could discuss about this.

I tested it over some datasets I created, but need some help about how to set up unit tests here.


timyitong:
Oops. I think someone has already submitted another pull request solving the same problem. Please ignore this one.

PR #14, opened 2015-01-21
Support LONG type in spark-avro

jonbender:


jonbender:
Whoops, redundant with #5

PR #15, opened 2015-01-21
Support globbed paths in AvroRelation

jonbender:


vlyubin:
Thanks for contributing Jon!

PR #16, opened 2015-01-21
Adds ability to save schemaRDDs as avro file

vlyubin:
If anyone knows how can I combine convertTypeToAvro and convertFieldTypeToAvro into one, let me know. They have almost identical code, but the builder classes are different and they don't have a common parent.


src/main/scala/com/databricks/spark/avro/AvroSaver.scala
```diff
@@ -0,0 +1,209 @@
...
+import org.apache.avro.generic.GenericRecord
+import org.apache.avro.mapred._
+import org.apache.avro.SchemaBuilder.{RecordBuilder, FieldAssembler}
+import org.apache.avro.{SchemaBuilder, Schema}
+
+import org.apache.hadoop.io.NullWritable
+import org.apache.hadoop.mapred.JobConf
+
+object AvroSaver {
+  /* This object provides a save() method that is used to save SchemaRDD as avro file at a
```
aarondav: Change to Javadoc (/** */, put before object definition) and, I haven't read it all yet, but try to make it more concise if possible, or at least introduce whitespace.

src/main/scala/com/databricks/spark/avro/AvroSaver.scala
```diff
@@ -0,0 +1,209 @@
...
+   * in sparkSQL doesn't have a name associated with it, we are taking the name of the last
+   * structure field that the current structure belongs to (in sparkSQL fields have a name, just as
+   * they do in avro). For example if the row at the top level had a field called "Auto", which
+   * happens to be a structure describing a car, we would call that structure "Auto". Note that in
+   * * case of schema like the following: Structure("A":1, "B": Array(Structure("X":1))), the inner
+   * * structure will get a name "B". When we process original rows, they get a name
+   * "topLevelRecord" (since we can't apply the rule above to them).
+   */
+
+  private def convertTypeToAvro[T](dataType: DataType,
```
aarondav: nit: putting all arguments at four-space indentation on their own line is more typical than aligning with the method name. i.e.,

```scala
pricate def convertTypeToAvro[T](
    dataType: DataType,
    ...
    ...): T = {
  dataType match {
  }
}
```

src/main/scala/com/databricks/spark/avro/AvroSaver.scala
```diff
@@ -0,0 +1,209 @@
...
+        item.asInstanceOf[Timestamp].getTime
+
+      case ArrayType(_, _) =>
+        val sourceArray = item.asInstanceOf[Seq[Any]]
+        val targetArray = new Array[Any](sourceArray.size)
+        var idx = 0
+
+        while (idx < sourceArray.size) {
+          targetArray(idx) = convertValueToAvro(sourceArray(idx),
+            dataType.asInstanceOf[ArrayType].elementType, structName)
```
aarondav: I think the elementType is the first parameter of ArrayType's unapply, so perhaps `case ArrayType(elementType, _)` would work.

src/main/scala/com/databricks/spark/avro/AvroSaver.scala
```diff
@@ -0,0 +1,209 @@
...
+          idx += 1
+        }
+
+        targetArray
+
+      case MapType(StringType, _, _) =>
```
aarondav: What happens to other kinds of Maps? Would they just get an exception when writing to avro?

src/main/scala/com/databricks/spark/avro/AvroSaver.scala
```diff
@@ -0,0 +1,209 @@
...
+        val javaMap = new HashMap[String, Any]()
+        item.asInstanceOf[Map[String, Any]].foreach(kv => javaMap.put(kv._1,
```
aarondav: typically we use { and put all code on a new line if the lambda is not a one-liner. e.g.,

```scala
item.asInstanceOf[Map[String, Any]].foreach { kv =>
  javaMap.put(kv._1, convertValueToAvro(kv._2, valueType, structName)))
}
javaMap
```

(That's assuming you use the valueType from the case statement as well.)

src/main/scala/com/databricks/spark/avro/AvroSaver.scala
vlyubin: When we are constructing schema and we encounter a map with non-string key type, we trigger "Unexpected type" error.

When converting actual items, avro schema has been created already, so all types should be correct (if there was an error, it should have been caught before). So if in this match statement dataType is a map, it keys must be strings.

src/main/scala/com/databricks/spark/avro/AvroSaver.scala
```diff
@@ -0,0 +1,205 @@
...
+      schemaBuilder: SchemaBuilder.BaseTypeBuilder[T],
+      structName: String): T = {
+    dataType match {
+      case ByteType => schemaBuilder.bytesType()
+      case ShortType => schemaBuilder.intType()
+      case IntegerType => schemaBuilder.intType()
+      case LongType => schemaBuilder.longType()
+      case FloatType => schemaBuilder.floatType()
+      case DoubleType => schemaBuilder.doubleType()
+      case DecimalType() => schemaBuilder.stringType()
```
aarondav: I don't think this unapply exists in the newest version of Spark, and it would be unclear what it means if it did (since decimal may have a fixed precision). Let's just match using the type: `case _: DecimalType` here and everywhere else that uses DecimalType.

src/main/scala/com/databricks/spark/avro/AvroSaver.scala
```diff
@@ -0,0 +1,205 @@
...
+ * avro types. One remark worth mentioning is the structName parameter that functions have. Avro
+ * records have a name associated with them, which must be unique. Since SturctType in sparkSQL
+ * doesn't have a name associated with it, we are taking the name of the last structure field that
+ * the current structure is a child of. For example if the row at the top level had a field called
+ * "X", which happens to be a structure, we would call that structure "X". When we process original
+ * rows, they get a name "topLevelRecord".
+ */
+object AvroSaver {
+
+  private def convertTypeToAvro[T](dataType: DataType,
```
aarondav: nit, should be:

```scala
private def convertTypeToAvro[T](
    dataType: DataType,
    schemaBuilder: SchemaBuilder.BaseTypeBuilder[T],
    structName: String): T = {
```

src/main/scala/com/databricks/spark/avro/AvroSaver.scala
```diff
@@ -0,0 +1,205 @@
...
+      case DecimalType() => schemaBuilder.stringType()
+      case StringType => schemaBuilder.stringType()
+      case BinaryType => schemaBuilder.bytesType()
+      case BooleanType => schemaBuilder.booleanType()
+      case TimestampType => schemaBuilder.longType()
+
+      case ArrayType(elementType, _) =>
+        val builder = SchemaBuilder.builder()
+        val elementSchema = convertTypeToAvro(elementType, builder, structName)
+
```
aarondav: probably don't need this extra space, here and for MapType()

src/main/scala/com/databricks/spark/avro/AvroSaver.scala
```diff
@@ -0,0 +1,205 @@
...
+        schemaBuilder.array().items(elementSchema)
+
+      case MapType(StringType, valueType, _) =>
+        val builder = SchemaBuilder.builder()
+        val valueSchema = convertTypeToAvro(valueType, builder, structName)
+
+        schemaBuilder.map().values(valueSchema)
+
+      case StructType(_) =>
+        convertStructToAvro(dataType.asInstanceOf[StructType],
```
aarondav: If you need to access the typed variable, you can do this funny syntax:

```scala
case structType @ StructType(_) =>
  convertStructToAvro(structType, schemaBuilder.record(structName))
```

However, since you're not actually using the pattern-match, it may be easier to read to just use
```scala
case structType: StructType =>
  convertStructToAvro(structType, schemaBuilder.record(structName))
```

src/main/scala/com/databricks/spark/avro/AvroSaver.scala
```diff
@@ -0,0 +1,205 @@
...
+          schemaBuilder.record(structName))
+
+      case other => sys.error(s"Unexpected type $dataType")
+    }
+  }
+
+  private def convertFieldTypeToAvro[T](dataType: DataType,
```
aarondav: see arguments nit above

src/main/scala/com/databricks/spark/avro/AvroSaver.scala
```diff
@@ -0,0 +1,205 @@
...
+      newFieldBuilder: SchemaBuilder.FieldTypeBuilder[T],
+      structName: String): SchemaBuilder.FieldDefault[_, _] = {
```
aarondav: just want to confirm that neither of these type parameters to FieldDefault should be T

src/main/scala/com/databricks/spark/avro/AvroSaver.scala
```diff
@@ -0,0 +1,205 @@
...
+      case DecimalType() => newFieldBuilder.stringType()
+      case StringType => newFieldBuilder.stringType()
+      case BinaryType => newFieldBuilder.bytesType()
+      case BooleanType => newFieldBuilder.booleanType()
+      case TimestampType => newFieldBuilder.longType()
+
+      case ArrayType(elementType, _) =>
+        val builder = SchemaBuilder.builder()
+        val elementSchema = convertTypeToAvro(elementType, builder, structName)
+
```
aarondav: same comment regarding extra space, just not needed for a 3-liner

src/main/scala/com/databricks/spark/avro/AvroSaver.scala
```diff
@@ -0,0 +1,205 @@
...
+        newFieldBuilder.array().items(elementSchema)
+
+      case MapType(StringType, valueType, _) =>
+        val builder = SchemaBuilder.builder()
+        val valueSchema = convertTypeToAvro(valueType, builder, structName)
+
+        newFieldBuilder.map().values(valueSchema)
+
+      case StructType(_) =>
```
aarondav: same comment on accessing typed variable

src/main/scala/com/databricks/spark/avro/AvroSaver.scala
```diff
@@ -0,0 +1,205 @@
...
+        convertStructToAvro(dataType.asInstanceOf[StructType],
+          newFieldBuilder.record(structName))
+
+      case other => sys.error(s"Unexpected type $dataType")
```
aarondav: I like to throw better exceptions when possible; IllegalArgumentException or IllegalStateException would be preferable to the RuntimeException thrown here.

src/main/scala/com/databricks/spark/avro/AvroSaver.scala
```diff
@@ -0,0 +1,205 @@
...
+    }
+  }
+
+  private def convertStructToAvro[T](structType: StructType, schemaBuilder: RecordBuilder[T]): T = {
+    val fieldsAssembler: FieldAssembler[T] = schemaBuilder.fields()
+    structType.fields foreach { field =>
```
aarondav: avoid infix, use .

src/main/scala/com/databricks/spark/avro/AvroSaver.scala
```diff
@@ -0,0 +1,205 @@
...
+      val newField = fieldsAssembler.name(field.name).`type`()
+
+      if (field.nullable) {
```
aarondav: I'm not 100% clear on how the avro type and methods related to each other. Would you mind making a picture (in text or some program) on which methods call which? Or at least, definitions of "field type", "type", "struct", and "value".

For instance, it is hard for me to verify that nullable fields always result in union types.

src/main/scala/com/databricks/spark/avro/AvroSaver.scala
```diff
@@ -0,0 +1,205 @@
...
+
+    while (fieldIterator.hasNext) {
+      val field = fieldIterator.next
+      record.put(field.name, convertValueToAvro(rowIterator.next, field.dataType, field.name))
+    }
+
+    (new AvroKey(record), NullWritable.get())
+  }
+
+  def save(schemaRDD: SchemaRDD, location: String): Unit = {
```
aarondav: This is a personal preference, but I like a source file to be laid out such that it can be read from top to bottom, as much as is feasible. I think you have basically done this, but inverted. If you prefer this style, that's fine, but if you have no preference, I think it'd be nice to reorder them such that a reader can start at this method and then basically just go down the file.

src/main/scala/com/databricks/spark/avro/AvroSaver.scala
```diff
@@ -0,0 +1,205 @@
...
+      case MapType(StringType, valueType, _) =>
+        val javaMap = new HashMap[String, Any]()
+        item.asInstanceOf[Map[String, Any]].foreach { case (key, value) =>
+          javaMap.put(key, convertValueToAvro(value, valueType, structName))
+        }
+        javaMap
+
+      case StructType(_) => rowToAvro(item.asInstanceOf[Row], dataType.asInstanceOf[StructType],
+        structName)._1.datum()
+    }
```
aarondav: add a `case other` like for the other guys to throw a nicer exception than MatchError

src/main/scala/com/databricks/spark/avro/AvroSaver.scala
```diff
@@ -0,0 +1,205 @@
...
+      } else {
+        convertFieldTypeToAvro(field.dataType, newField, field.name).noDefault()
+      }
+    }
+    fieldsAssembler.endRecord()
+  }
+
+  private def convertValueToAvro(item: Any, dataType: DataType, structName: String): Any = {
+    if (item == null) {
+      // Since some schema fields might be nullable, item can be null. If so, just return it
```
aarondav: Should we ensure the data type is nullable (i.e., throw exception here otherwise)? If it's not, then we may risk writing invalid data to Avro, if it's not specifically a union type. I'm not sure of the behavior in this scenario.

src/main/scala/com/databricks/spark/avro/AvroSaver.scala
```diff
@@ -0,0 +1,205 @@
...
+        }
+        javaMap
+
+      case StructType(_) => rowToAvro(item.asInstanceOf[Row], dataType.asInstanceOf[StructType],
+        structName)._1.datum()
+    }
+  }
+
+  private def rowToAvro(row: Row, structType: StructType,
+      structName: String):(AvroKey[GenericRecord], NullWritable) = {
```
aarondav: nit: same arguments-on-new-line thing and space before colon

src/main/scala/com/databricks/spark/avro/AvroSaver.scala
vlyubin: Avro will throw an Exception that you're trying to write a wrong data type.

src/main/scala/com/databricks/spark/avro/AvroSaver.scala
vlyubin: Can I add a comment that at this point we've verified schema correctness and there can be no other type?

src/main/scala/com/databricks/spark/avro/AvroSaver.scala
vlyubin: Ok, I can reorder stuff. I do have a preference of writing helper methods at the top and main functions at the bottom, but that's a result of writing many one file C++ programs,

src/main/scala/com/databricks/spark/avro/AvroSaver.scala
aarondav: Sure, but should probably handle it in the same way in each method. Throwing nice exceptions is good when something unanticipated happens.

src/main/scala/com/databricks/spark/avro/AvroSaver.scala
vlyubin: After the new commit, this actually became simpler than it was before. Can you let me know if you still need a document with which methods call which.

P.S. I was not considering nullable map values / array elements before, fixed now.

src/main/scala/com/databricks/spark/avro/AvroSaver.scala
```diff
@@ -0,0 +1,214 @@
...
+ * avro types. One remark worth mentioning is the structName parameter that functions have. Avro
+ * records have a name associated with them, which must be unique. Since SturctType in sparkSQL
+ * doesn't have a name associated with it, we are taking the name of the last structure field that
+ * the current structure is a child of. For example if the row at the top level had a field called
+ * "X", which happens to be a structure, we would call that structure "X". When we process original
+ * rows, they get a name "topLevelRecord".
+ */
+object AvroSaver {
+
+  def save(schemaRDD: SchemaRDD, location: String): Unit = {
```
aarondav: it looks like this:
```

save ---> rowToAvro ---> structToAvro ---> fieldTypeToAvro ---> typeToAvro
          \    \                  \------<------/-------<-------/  /-<-\
           \    \   
            \    \-----> valueToAvro
             -----<-----/    /-<-\
```

or, with annotated edges:
```
     per_row       schema             per_field          array/map
save ---> rowToAvro ---> structToAvro ---> fieldTypeToAvro ---> typeToAvro
          \    \                  \------<------/-------<-------/  /-<-\
           \    \per_col               struct         struct     array/map
            \    \-----> valueToAvro
             -----<-----/    /-<-\
                struct    array/map
```

aarondav: Ahh, I see, the struct/fieldtype/type stuff is just for the schema, is that the case?

vlyubin:
Benchmarks of running .count()

// 1GB avro file with 800K entries. Those entires contain variery of datatypes, including large maps, strings, arrays and nested record:

Old way: 18.75 | 14.31 | 14.30 | 14.71 | 14.25 | 13.36 | 14.03 | 14.88 (average = 14.82)
New way: 15.50 | 10.36 | 10.29 | 10.09 | 10.92 | 10.59 | 9.92  | 10.03 (average = 10.96)

// 4.3 GB file with 2M entries, but less variety of data types

Old way: 52.08 | 45.46 | 45.69 (average = 47.74)
New way: 40.99 | 41.02 | 41.06 (average = 41.02)

// 4 files x 4.3 GB each

Old way: 186.58 | 179.68 | 179.14 (average = 181.8)
New way: 167.55 | 163.98 | 162.72 (average = 164.75)

src/main/scala/com/databricks/spark/avro/AvroSaver.scala
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  private def rowsToAvro(
+      rows: Iterator[Row],
+      schema: StructType): Iterator[(AvroKey[GenericRecord], NullWritable)] = {
+    val converter = createConverter(schema, "topLevelRecord")
+    val records = ArrayBuffer[(AvroKey[GenericRecord], NullWritable)]()
+
+    while (rows.hasNext) {
+      records.+=((new AvroKey(converter(rows.next).asInstanceOf[Record]), NullWritable.get()))
```
aarondav: for collections-y things, it's fine to use infix notation -- `records += new AvroKey(...)`

src/main/scala/com/databricks/spark/avro/SchemaConverters.scala
```diff
@@ -0,0 +1,185 @@
...
+        newFieldBuilder.map().values(valueSchema)
+
+      case structType: StructType =>
+        convertStructToAvro(structType, newFieldBuilder.record(structName))
+
+      case other => throw new IllegalArgumentException(s"Unexpected type $dataType.")
+    }
+  }
+
+  private def getBuilder(isNullable: Boolean) = {
```
aarondav: maybe rename to createSchemaBuilder or getSchemaBuilder and add an explicit return type

src/main/scala/com/databricks/spark/avro/SchemaConverters.scala
```diff
@@ -0,0 +1,185 @@
...
+        schemaBuilder.map().values(valueSchema)
+
+      case structType: StructType =>
+        convertStructToAvro(structType, schemaBuilder.record(structName))
+
+      case other => throw new IllegalArgumentException(s"Unexpected type $dataType.")
+    }
+  }
+
+  private def convertFieldTypeToAvro[T](
```
aarondav: Please add brief method comments for this method and convertTypeToAvro -- i.e., that this is used for fields of records, and type is used for everything else.

src/main/scala/com/databricks/spark/avro/SchemaConverters.scala
```diff
@@ -0,0 +1,185 @@
...
+import org.apache.spark.sql._
+import org.apache.spark.sql.catalyst.types.{MapType, ArrayType}
+import org.apache.avro.Schema.Type._
+
+
+/**
+ * This object contains method that are used to convert sparkSQL schemas to avro schemas and vice
+ * versa.
+ */
+object SchemaConverters {
```
aarondav: This guy looks very clean, thanks for doing this refactor!

src/main/scala/com/databricks/spark/avro/AvroRelation.scala
```diff
@@ -120,56 +81,102 @@ case class AvroRelation(location: String)(@transient val sqlContext: SQLContext)
     DataFileReader.openReader(input, reader)
   }
 
-  private case class SchemaType(dataType: DataType, nullable: Boolean)
-
-  private def toSqlType(avroSchema: Schema): SchemaType = {
-    import Schema.Type._
+  def createConverter(schema: Schema): (Any) => Any = {
```
aarondav: Make private and add a method comment, especially about the return type.

src/main/scala/com/databricks/spark/avro/AvroRelation.scala
```diff
@@ -46,44 +50,6 @@ case class AvroRelation(location: String)(@transient val sqlContext: SQLContext)
     convertedSchema
   }
 
-  /** 
-   * There are several properties of Avro that we have to account for in this method:
-   * 1) Avro uses Utf8 strings, so we want to convert them to java.lang.String for SparkSQL,
```
aarondav: Should probably keep these comments in the moved method doc.

aarondav: Ah, I guess only the avro map keys thing applies to the moved version of this method. Perhaps just add inline comments for 2 and 3 where appropriate -- I see 1 already has one.

src/main/scala/com/databricks/spark/avro/AvroSaver.scala
aarondav: Also I think it's reasonable to use `rows.map` here on the virtue of iterators being lazily computed and not requiring a whole partition to fit in memory (and writing an iterator from scratch is more code than it's worth to get around the Scala overhead).

src/main/scala/com/databricks/spark/avro/AvroSaver.scala
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  /**
+   * This function constructs converter function for a given sparkSQL datatype. These functions
+   * will be used to convert schemaRDD to avro format.
+   */
+  def createConverter(dataType: DataType, structName: String): (Any) => Any = {
+    dataType match {
+      case ByteType | ShortType | IntegerType | LongType | FloatType | DoubleType | StringType |
+           BinaryType | BooleanType =>
```
aarondav: Just want to confirm that your test includes a BinaryType, because it seems to be a ByteBuffer coming in, but this would probably be an Array[Byte] going out -- perhaps Avro handles that correctly, I just want to be sure.

src/main/scala/com/databricks/spark/avro/AvroSaver.scala
```diff
@@ -0,0 +1,146 @@
...
+        val elementConverter = createConverter(elementType, structName)
+
+        (item: Any) => {
+          if (item == null) return null
+
+          val sourceArray = item.asInstanceOf[Seq[Any]]
+          val targetArray = new Array[Any](sourceArray.size)
+          var idx = 0
+
+          while (idx < sourceArray.size) {
```
aarondav: Seq's size method is usually O(n), let's save the size in a variable. (This has historically been the cause of a several perf problems, spinning on computing  the size of Seq.)

src/main/scala/com/databricks/spark/avro/AvroSaver.scala
```diff
@@ -0,0 +1,146 @@
...
+      case TimestampType =>
+        (item: Any) => {
+          if (item == null) null else item.asInstanceOf[Timestamp].getTime
+        }
+
+      case ArrayType(elementType, _) =>
+        val elementConverter = createConverter(elementType, structName)
+
+        (item: Any) => {
+          if (item == null) return null
```
aarondav: This return is probably converted into a `throw new NonLocalControlThrowable` (which is how Scala implements returns within closures). This could be very expensive in some situations; I would convert it to an `if { } else { }` to be safe.

src/main/scala/com/databricks/spark/avro/AvroSaver.scala
```diff
@@ -0,0 +1,146 @@
...
+
+          val javaMap = new HashMap[String, Any]()
+          item.asInstanceOf[Map[String, Any]].foreach { case (key, value) =>
+            javaMap.put(key, valueConverter(value))
+          }
+          javaMap
+        }
+
+      case StructType(_) =>
```
aarondav: Can write this `structType: StructType` and not do the cast below, I believe.

aarondav:
Looking pretty good! I think this is close to merging.

src/main/scala/com/databricks/spark/avro/AvroSaver.scala
vlyubin: Wow, that's good to know!

src/main/scala/com/databricks/spark/avro/AvroSaver.scala
vlyubin: test("conversion to avro and back") was capturing this, but I added an extra test.

aarondav:
LGTM.

@marmbrus Would you mind taking a brief look at AvroSaver and SchemaConverters?

src/main/scala/com/databricks/spark/avro/SchemaConverters.scala
```diff
@@ -0,0 +1,194 @@
...
+
+      case other => sys.error(s"Unsupported type $other")
+    }
+  }
+
+  /**
+   * This function converts sparkSQL StructType into avro schema. This method uses two other
+   * converter methods in order to do the conversion.
+   */
+  def convertStructToAvro[T](structType: StructType, schemaBuilder: RecordBuilder[T]): T = {
```
marmbrus: private?

src/main/scala/com/databricks/spark/avro/SchemaConverters.scala
```diff
@@ -0,0 +1,194 @@
...
+import org.apache.spark.sql._
+import org.apache.spark.sql.catalyst.types.{MapType, ArrayType}
+import org.apache.avro.Schema.Type._
+
+
+/**
+ * This object contains method that are used to convert sparkSQL schemas to avro schemas and vice
+ * versa.
+ */
+object SchemaConverters {
```
marmbrus: actually should this who class be private?

marmbrus:
LGTM

src/main/scala/com/databricks/spark/avro/SchemaConverters.scala
vlyubin: This is the method that I actually call from AvroSaver for each row. So I can't make it private.

src/main/scala/com/databricks/spark/avro/SchemaConverters.scala
vlyubin: You're right, this object is only intended for use by AvroRelation and AvroSaver. I'll make it private.

src/main/scala/com/databricks/spark/avro/SchemaConverters.scala
```diff
@@ -0,0 +1,194 @@
...
+import org.apache.spark.sql._
+import org.apache.spark.sql.catalyst.types.{MapType, ArrayType}
+import org.apache.avro.Schema.Type._
+
+
+/**
+ * This object contains method that are used to convert sparkSQL schemas to avro schemas and vice
+ * versa.
+ */
+private object SchemaConverters {
```
aarondav: Note that you can always make things `private[avro]` so that you can access them internally but not expose them to users (e.g., through javadoc)

src/main/scala/com/databricks/spark/avro/SchemaConverters.scala
marmbrus: `private[avro]`?

aarondav:
Looks like this is conflicting with another PR that was merged -- please rebase and we'll merge.

src/main/scala/com/databricks/spark/avro/SchemaConverters.scala
vlyubin: Oh, OK.

aarondav:
Great, merging!

PR #18, opened 2015-01-30
Stuff for benchmark testing + README changes

vlyubin:


README.md
```diff
@@ -69,15 +69,31 @@ fixed -> BinaryType
...
+DecimalType -> string
+BinaryType -> bytes
+TimestampType -> long
+StructType -> record
+```
+
 ### Scala API
 
-You can use the library by loading the implicits from `com.databricks.spark.avro._`.
+To get SchemaRDD from avro file, you can use the library by loading the implicits from `com.databricks.spark.avro._`.
```
aarondav: nit: to get *a* SchemaRDD (also, this will have to be updated once the data frame change goes in, just as a side note)

README.md
```diff
@@ -121,4 +143,9 @@ JavaSchemaRDD episodes = AvroUtils.avroFile(sqlContext, "episodes.avro");
 ```
 
 ## Building From Source
-This library is built with [SBT](http://www.scala-sbt.org/0.13/docs/Command-Line-Reference.html), which is automatically downloaded by the included shell script.  To build a JAR file simply run `sbt/sbt package` from the project root. To run the tests, you should run `sbt/sbt test`.
+This library is built with [SBT](http://www.scala-sbt.org/0.13/docs/Command-Line-Reference.html), which is automatically downloaded by the included shell script.  To build a JAR file simply run `sbt/sbt package` from the project root.
+
+## Testing
+To run the tests, you should run `sbt/sbt test`. In case you are doing improvements that target speed, you can generate a sample avro file and check how long does it take to read that avro file using the following commands:
+`sbt/sbt "test:run-main com.databricks.spark.avro.AvroFileGenerator NUMBER_OF_RECORDS NUMBER_OF_FILES"` will create sample avro files in `src/test/resources/avroForBenchmark/`. You can specify the number of records for each file, as well as the overall number of files.
```
aarondav: we should probably generate this file in `target/` instead of `src/`, or at least add the directory to the `.gitignore`.

src/test/scala/com/databricks/spark/avro/AvroFileGenerator.scala
```diff
@@ -0,0 +1,96 @@
...
+
+  def generateRandomMap(rand: Random): java.util.Map[String, Int] = {
+    val jMap = new HashMap[String, Int]()
+    for (i <- 0 until objectSize) {
+      jMap.put(rand.nextString(5), i)
+    }
+    jMap
+  }
+  
+  def generateRandomArray(rand: Random): Vector[Boolean] = {
```
aarondav: Any particular reason you used util.Vector? util.ArrayList is more common unless you explicitly need the synchronization advantages of the former.

src/test/scala/com/databricks/spark/avro/AvroFileGenerator.scala
```diff
@@ -0,0 +1,96 @@
...
+
+    if (args.size > 0) {
+      numberOfRecords = args(0).toInt
+    }
+
+    if (args.size > 1) {
+      numberOfFiles = args(1).toInt
+    }
+
+    new File(outputDir).mkdir() // Create directory for output files
```
aarondav: Might as well switch this to mkdirs, I think, especially if we move it to target.

src/test/scala/com/databricks/spark/avro/AvroFileGenerator.scala
```diff
@@ -0,0 +1,96 @@
...
+  def main(args: Array[String]) {
+    var numberOfRecords = defaultNumberOfRecords
+    var numberOfFiles = defaultNumberOfFiles
+
+    if (args.size > 0) {
+      numberOfRecords = args(0).toInt
+    }
+
+    if (args.size > 1) {
+      numberOfFiles = args(1).toInt
```
aarondav: Maybe add a println for the final numbers chosen (help users who may confuse the argument order).

src/test/scala/com/databricks/spark/avro/AvroFileGenerator.scala
```diff
@@ -0,0 +1,96 @@
...
+      avroRec.put("bytes", generateRandomByteBuffer(rand))
+
+      dataFileWriter.append(avroRec)
+      idx += 1
+    }
+
+    dataFileWriter.close()
+  }
+
+  def main(args: Array[String]) {
```
aarondav: Should we delete old files in the target directory?

src/test/scala/com/databricks/spark/avro/AvroReadBenchmark.scala
```diff
@@ -0,0 +1,35 @@
...
+ * This object runs a simple benchmark test on the avro files in benchmarkFilesDir. It measures
+ * how long does it take to convert them into SchemaRDD and run count() method on them. See
+ * README on how to invoke it.
+ */
+object AvroReadBenchmark {
+
+  val benchmarkFilesDir = "src/test/resources/avroForBenchmark/"
+
+  def main(args: Array[String]) {
+    if (new File(benchmarkFilesDir).list.isEmpty) {
```
aarondav: list may return null if the dir does not exist, so I would add a null check first.

src/test/scala/com/databricks/spark/avro/AvroReadBenchmark.scala
```diff
@@ -0,0 +1,35 @@
...
+      sys.error("First you should generate some files to run a benchmark with (see README)")
```
aarondav: Maybe add a line saying what directory was not discovered.

src/test/scala/com/databricks/spark/avro/AvroReadBenchmark.scala
```diff
@@ -0,0 +1,35 @@
...
+
+
+/**
+ * This object runs a simple benchmark test on the avro files in benchmarkFilesDir. It measures
+ * how long does it take to convert them into SchemaRDD and run count() method on them. See
+ * README on how to invoke it.
+ */
+object AvroReadBenchmark {
+
+  val benchmarkFilesDir = "src/test/resources/avroForBenchmark/"
```
aarondav: Can probably just use AvroFileGenerator.outputPath

src/test/scala/com/databricks/spark/avro/AvroReadBenchmark.scala
```diff
@@ -0,0 +1,35 @@
...
+
+  def main(args: Array[String]) {
+    if (new File(benchmarkFilesDir).list.isEmpty) {
+      sys.error("First you should generate some files to run a benchmark with (see README)")
+    }
+
+    println("\n\n\nStaring benchmark test - creating SchemaRDD from benchmark avro files\n\n\n")
+
+    val startTime = System.currentTimeMillis
```
aarondav: nit: prefer nanoTime to currentTimeMillis; the latter is not necessarily monotonic (e.g., ntp)

aarondav:
LGTM, only superficial comments.

README.md
aarondav: update the path above to `target/avroForBenchmark`

src/test/scala/com/databricks/spark/avro/AvroFileGenerator.scala
```diff
@@ -0,0 +1,103 @@
...
+      numberOfRecords = args(0).toInt
+    }
+
+    if (args.size > 1) {
+      numberOfFiles = args(1).toInt
+    }
+
+    println(s"Generating $numberOfFiles avro files with $numberOfRecords records each")
+
+    FileUtils.deleteDirectory(new File(outputDir)) // Delete contents of output directory
```
aarondav: We don't directly depend on apache.commons, so we probably shouldn't call methods on it (to avoid version-related issues). Let's just do something like

```scala
new File(outputDir).listFiles().filter(_.isFile).foreach(_.delete())
```

aarondav:
Merging, thanks!

PR #20, opened 2015-02-03
Test for globbed paths support

vlyubin:


aarondav:
LGTM. I wonder what happens if you do `src/*/*/{episodes.avro,episodes.avro}`.

vlyubin:
@aarondav `src/*/*/{episodes.avro,episodes.avro}` works as well.

aarondav:
But does it duplicate the data?

vlyubin:
No, it doesn't (`collect()` still returns 8 entries). This is reasonable though, right? `src/*/*/*/{test.avro,test2.avro}` gets both of them as expected.

aarondav:
Yup, 8 is what I would have expected, sounds goodl.

PR #22, opened 2015-02-04
Write benchmarking and minor changes to test utilities

vlyubin:


src/test/scala/com/databricks/spark/avro/AvroSuite.scala
```diff
@@ -163,14 +172,14 @@ class AvroSuite extends FunSuite {
...
     }
 
-    FileUtils.deleteDirectory(tempDir)
+    DirectoryDeletion.recursiveDelete(tempDir)
   }
 
   test("converting some specific sparkSQL types to avro") {
     val testSchema = StructType(Seq(StructField("Name", StringType, false),
-                                    StructField("Length", IntegerType, false),
+                                    StructField("Length", IntegerType, true),
```
vlyubin: Nullable items are already a part of the previous test ("conversion to avro and back"), but I want this test to have nullable fields as well.

src/test/scala/com/databricks/spark/avro/AvroSuite.scala
```diff
@@ -29,6 +29,15 @@ import org.scalatest.FunSuite
 /* Implicits */
 import TestSQLContext._
 
+private[avro] object DirectoryDeletion {
```
vlyubin: What I needed was a method to recursively delete directories so that we don't use FileUtils. This method is used by AvroSuite, WriteBenchmark and AvroFileGenerator. Since all of those are a part of testing infrastructure, AvroSuite would be a good place to put it. I would have made it a static method, but since I can't, I just have a separate object that has this method. Is this considered hacky?

aarondav: No, not hacky -- this is pretty common for libraries, which have to minimize their dependencies. I would create an object called TestUtils and put this in there.

aarondav: But I might actually copy Spark's implementation of the same thing from here: https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/util/Utils.scala#L908 (remove the shutdownDeletePaths thing), with a note of where it came from and after putting it in its own utility class.

src/test/scala/com/databricks/spark/avro/AvroSuite.scala
```diff
@@ -29,6 +29,15 @@ import org.scalatest.FunSuite
...
+  private[avro] def recursiveDelete(path: File): Unit = {
+    if (path.isDirectory) {
+      path.listFiles.foreach(recursiveDelete(_))
+    }
+    path.delete
```
aarondav: fyi: use of method invocation without parentheses is generally restricted to non-side-effect-y things, such as getters. Otherwise, it can hide method invocation versus value accessing. In general, when unsure, I would lean towards using parentheses.

src/test/scala/com/databricks/spark/avro/AvroSuite.scala
```diff
@@ -16,20 +16,90 @@
...
+          // Delete can also fail if the file simply did not exist
+          if (file.exists()) {
+            throw new IOException("Failed to delete: " + file.getAbsolutePath)
+          }
+        }
+      }
+    }
+  }
+
+  // This function generates a random map(string, int) of a given size.
```
aarondav: nit: javadoc comments instead

src/test/scala/com/databricks/spark/avro/AvroWriteBenchmark.scala
```diff
@@ -0,0 +1,73 @@
...
+
+
+/**
+ * This object runs a simple benchmark test to find out how long does it take to write a large
+ * SchemaRDD to an avro file. It reads one argument, which specifies how many rows does the
+ * SchemaRDD that we're writing contain.
+ */
+object AvroWriteBenchmark {
+
+  val defaultNumberOfRows = 50000
```
aarondav: About how long does this take? Should probably make it big enough to take around a minute or more, by default.

aarondav:
LGTM

src/test/scala/com/databricks/spark/avro/AvroWriteBenchmark.scala
vlyubin: 50K it taking around 50 seconds. Increasing it to 100K / 70K results in OutOfMemoryError on my machine, unless I pass an option to increase heap size.
Technically 50K is still enough to tell the difference and if there a need, one can always pass a custom number of rows.

src/test/scala/com/databricks/spark/avro/AvroSuite.scala
vlyubin: Ok, changed.

src/test/scala/com/databricks/spark/avro/AvroWriteBenchmark.scala
aarondav: Oh, it should not throw an OOM based on size, that just made me realize we're not generating the rows sanitarily. We should be generating them like

```scala
sc.parallelize(0 until numberOfRows).map(_ => generateRandomRow())
```

(really we shouldn't even be parallelizing that range, but that should be fine for numberOfRows < a billion)

aarondav: (This is especially important because it suggests we may not really testing the performance of the saver, but just Spark in shipping around large byte arrays and dealing with GC pauses.)

src/main/scala/com/databricks/spark/avro/AvroRelation.scala
```diff
@@ -167,7 +167,7 @@ case class AvroRelation(location: String)(@transient val sqlContext: SQLContext)
             null
           } else {
             // Avro map keys are always strings, so it's enough to just call toString on them.
-            item.asInstanceOf[Map[Any, Any]].map(x => (x._1.toString, valueConverter(x._2)))
+            item.asInstanceOf[Map[Any, Any]].map(x => (x._1.toString, valueConverter(x._2))).toMap
```
vlyubin: This isn't quite related to this PR, but is sort of a bug fix. Apparently, before this change we would get a HashMap, and while it didn't cause any troubles for avro <-> sparkSQL, when I tried to read an avro file and save it as parquet, scala Map was expected. This fixes that. Let me know if I should make a separate PR for this though.

aarondav:
OK, final LGTM this time -- merging!

PR #25, opened 2015-02-05
Fix nullable fields

vlyubin:


src/main/scala/com/databricks/spark/avro/SchemaConverters.scala
```diff
@@ -101,10 +101,10 @@ private object SchemaConverters {
       val newField = fieldsAssembler.name(field.name).`type`()
 
       if (field.nullable) {
-        newField.nullable()
```
vlyubin: I mistakenly thought that this was changing the newField object, and there wasn't a test to catch this error.

src/test/scala/com/databricks/spark/avro/AvroSuite.scala
```diff
@@ -66,26 +65,26 @@ class AvroSuite extends FunSuite {
       .avroFile(testFile)
       .select('string)
       .collect()
-    assert(str.map(x => x(0)).toSet.contains("Terran is IMBA!"))
+    assert(str.map(_(0)).toSet.contains("Terran is IMBA!"))
```
vlyubin: :)

src/test/scala/com/databricks/spark/avro/AvroSuite.scala
```diff
@@ -137,9 +140,9 @@ class AvroSuite extends FunSuite {
 
     val tempDir = Files.createTempDir()
     val avroDir = tempDir + "/avro"
-    AvroSaver.save(TestSQLContext.avroFile(testToAvroFile), avroDir)
+    AvroSaver.save(TestSQLContext.avroFile(testFile), avroDir)
```
vlyubin: testToAvroFile was there for testing AvroSaver since the very beginning, because testFile had problems with nullables. Everything is OK now, so testFile can be used for both purposes.

aarondav:
LGTM

PR #28, opened 2015-02-09
Make avroFile reader has the option of choosing a partition number instead of using the default one

timyitong:
I hesitate a bit in which pattern to use for adding the option:
```
// Use Option
AvroRelation(location: String, minPartitions: Option[Int] = None)

// Just use a default value
AvroRelation(location: String, minPartitions: Int = 0)
```

I choose the latter one because we also have Java codes calling AvroRelation, and I feel it a less elegant to import scala None,Option classes into java codes to make it compile. Please suggest whether this choice makes sense.

I tested my change with an added unit tests and also in a Spark 1.2 shell where I was reading larger number of partitions of files in HDFS. The set-ups seem to be working but the final partitions number won't equal to the exact value of the parameter, but something a bit larger. For example, when I am reading 22 partitioned files, and choose the minPartitions to be 120. The final partitions size I get won't be 120, but will be k*22 where k is an integer making 22k just larger than 120 (in this example, k=6). I think this is expected as it behaves the same in Hadoop. But our unit tests only test one partition case, should I add a folder of small partitioned files and test this case?


timyitong:
Could anybody give some feedback about this change request?

src/main/java/com/databricks/spark/avro/AvroUtils.java
```diff
@@ -24,7 +24,13 @@
 public class AvroUtils {
     /** Returns a Schema RDD for the given avro path. */
     public static JavaSchemaRDD avroFile(JavaSQLContext sqlContext, String path) {
-        AvroRelation relation = new AvroRelation(path, sqlContext.sqlContext());
+        AvroRelation relation = new AvroRelation(path, 0, sqlContext.sqlContext());
+        return sqlContext.baseRelationToSchemaRDD(relation);
```
vlyubin: Maybe just use `return AvroUtils.avroFile (sqlContext, path, 0);` to avoid duplication.

vlyubin:
1) Using 0 instead of None sounds good to me.
2) " But our unit tests only test one partition case, should I add a folder of small partitioned files and test this case?" - if you can do it, that would be great! Add that folder to resources ...
3) If it's not too hard, can you squash your commits (e.g. there are two commits named "Code style changes", and it might be better to just combine them with the main commit (Add partitions number option for Avro reader))

I don't really have any other comments, so @aarondav will take a look at this.

timyitong:
@vlyubin : I generated 11 random files with 3 records each for testing the multiple partitions case. But I found the actual number of partitions in the result are even more arbitrary than I thought.

I test it against other large real datasets, and I think it still sticks to the partition number rule I mentioned previously.

But when the total number of records are not so large, and the sizes of records vary a lot (because it is randomly generated), there seems to be no obvious patterns how exactly the number of partitions would be in the end. I kinda suspect, Hadoop file reader decide which record goes to which partition based upon split.size, and when the record size varies too much, there would be some partitions that are wasted (cannot fit in one more record, but not yet close to the split.size).

As a result, I only verify the partitions number is larger than the number specified in the test case. Do you think this is okay? Of course, I can also generate more equal sized files, but that feels a bit awkward as we would be tweaking test inputs to make certain stuff pass.

src/main/scala/com/databricks/spark/avro/AvroRelation.scala
```diff
@@ -53,12 +53,15 @@ case class AvroRelation(location: String)(@transient val sqlContext: SQLContext)
 
   // By making this a lazy val we keep the RDD around, amortizing the cost of locating splits.
   override lazy val buildScan = {
+    val minPartitionsNum = if (minPartitions <= 0) sqlContext.sparkContext.defaultMinPartitions
```
aarondav: nit: ternary operator is only used for 1-liners. This should be expanded to:

```scala
if (minPartitions <= 0) {
  sqlContext.sparkContext.defaultMinPartitions
} else {
  minPartitions
}
```

aarondav:
LGTM, just one minor nit and I think we're good to go. The parameter is called "minPartitions", so I think the test is fine :)

timyitong:
Updated according to @aarondav 's feedback.

aarondav:
Merging, thanks!

PR #35, opened 2015-03-16
Change Record to GenericRecord

vlyubin:
Record implements GenericRecord. GenericRecord is enough in this scenario, and there are breaking scenarios for using Record here.


aarondav:
LGTM

vlyubin:
Thanks to @medale for finding this!

PR #41, opened 2015-03-31
Style and README cleanup.

rxin:
The Python code example in README was actually wrong (it was using // for comment).


vlyubin:
LGTM

PR #43, opened 2015-04-09
Respect AvroInputFormat file extension property

jaley:
As I mentioned in Issue 4, there's an `avro.mapred.ignore.inputs.without.extension` property used by `AvroInputFormat` to configure whether or not it should skip files that end in `.avro`. Currently, spark-avro ignores that and always skips files that don't have the extension. This change makes spark-avro consistent with the input format implementation, so if users set the property required for `AvroInputFormat`, spark-avro won't effectively override it anymore.

Relevant javadoc for the input format class:
https://avro.apache.org/docs/1.7.6/api/java/org/apache/avro/mapred/AvroInputFormat.html

### DIFF
--- a/src/main/scala/com/databricks/spark/avro/AvroRelation.scala
+++ b/src/main/scala/com/databricks/spark/avro/AvroRelation.scala
@@ -44,6 +44,8 @@ case class AvroRelation(
   extends BaseRelation with TableScan with InsertableRelation {
   var avroSchema: Schema = null
 
+  val IgnoreAvroFileExtensionProperty = "avro.mapred.ignore.inputs.without.extension"
+
   override val schema: StructType = {
     if (userSpecifiedSchema.isDefined) {
       // We need avroSchema to construct converter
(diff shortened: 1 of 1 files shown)


vlyubin:
LGTM, cc @aarondav

src/main/scala/com/databricks/spark/avro/AvroRelation.scala
```diff
@@ -105,11 +108,17 @@ case class AvroRelation(
       .toStream
       .map(_.getPath)
       .flatMap(getAllFiles(fs)(_))
+    val isAvroFile : Path => Boolean =
+      if(hadoopConfiguration.getBoolean(IgnoreAvroFileExtensionProperty, true)) {
+        _.getName.endsWith("avro")
+      } else {
+        Function.const(true)
+      }
     val singleFile = statuses
```
aarondav: maybe something like this (to avoid creating this temporary function)
```scala
val singleFile = if (hadoopConfiguration.getBoolean(IgnoreAvroFileExtensionProperty, true)) {
  statuses.find(_.getName.endsWith("avro"))
} else {
  statuses.headOption
}.getOrElse(sys.error(s"Could not find .avro file with schema at $path"))
```
also, consider renaming IgnoreAvroFileExtensionProperty, because that reads like the opposite of what it is

jaley:
@aarondav Thanks for comments - updated patch with your suggestions.

aarondav:
LGTM, merging into master. Thanks for the contribution!

PR #51, opened 2015-05-30
Alias and Doc support in Spark Schema

vasanthrj:
Preserves avro doc and aliases while save
Include avro alias column in DataFrame schema


vasanthrj:
Can someone review this? Please give some suggestion.

README.md
```diff
@@ -127,6 +127,14 @@ To save DataFrame as avro you should use the `save` method in `AvroSaver`. For e
 ```scala
 scala> AvroSaver.save(myRDD, "my/output/dir")
 ```
+
+To include aliases column in scheme invoke the method `addAvroAliasColumns()` of DataFrame.
+With alias while saving `saveAsAvroFile`, alias columns in DataFrame schema will not be include in Avro file.
+Alias will be available in `aliases` of avro schema.
```
jendap: Is it stored as "aliases" or "_aliases"?

src/main/scala/com/databricks/spark/avro/SchemaConverters.scala
```diff
@@ -98,14 +119,26 @@ private object SchemaConverters {
...
     structType.fields.foreach { field =>
-      val newField = fieldsAssembler.name(field.name).`type`()
-
-      if (field.nullable) {
-        convertFieldTypeToAvro(field.dataType, newField.nullable(), field.name, recordNamespace)
-          .noDefault
-      } else {
-        convertFieldTypeToAvro(field.dataType, newField, field.name, recordNamespace)
-          .noDefault
+      breakable {
```
jendap: Would it be more readable to just make it "if ... else" statement instead of "breakable" and "break"?

jendap:
I love the idea! And it seems to be working just fine.

A few suggestions...

1) Let's ask @marmbrus if "_doc" and "_aliases" are a good naming convention. It would be nice to make it consistent with other formats supporting it.

2) Is the implicit addAvroAliasColumns a good choice? It would be on every DataFrame - even after modifying columns and wiping out their metadata. Would it make sense to make it a property on SQLContext the same way as [parquet settings](http://spark.apache.org/docs/latest/sql-programming-guide.html#configuration) are done currently?

3) It would be nice to add an assert to the place where files are read after saving. Check that doc and aliases made the loop to disk and back.

4) Would it not be enough and more readable to add the doc and alias on one field in the test only? It does not have to be tested on all the different fields, right? It would save some 100 lines of code in tests ;-)

src/main/scala/com/databricks/spark/avro/SchemaConverters.scala
marmbrus: +1

src/main/scala/com/databricks/spark/avro/AvroSaver.scala
```diff
@@ -158,7 +158,12 @@ object AvroSaver {
 
             while (convertersIterator.hasNext) {
               val converter = convertersIterator.next()
-              record.put(fieldNamesIterator.next(), converter(rowIterator.next()))
+              val fieldName = fieldNamesIterator.next()
+              if(schema.getField(fieldName) != null) {
```
JoshRosen: Style nit: space after `if`.

src/main/scala/com/databricks/spark/avro/SchemaConverters.scala
```diff
@@ -49,7 +49,15 @@ private object SchemaConverters {
       case RECORD =>
         val fields = avroSchema.getFields.map { f =>
           val schemaType = toSqlType(f.schema())
-          StructField(f.name, schemaType.dataType, schemaType.nullable)
+          var meta = new MetadataBuilder()
+          if (f.doc != null) meta.putString("_doc", f.doc)
+          if(f.aliases() != null && f.aliases().size() > 0) {
```
JoshRosen: Style nit: space after `if`

src/main/scala/com/databricks/spark/avro/SchemaConverters.scala
```diff
@@ -88,6 +96,19 @@ private object SchemaConverters {
     }
   }
 
+  def dataFrameWithAliasColumn(df : DataFrame) : DataFrame = {
+    var newDf = df
+    for(field <- df.schema.fields) {
```
JoshRosen: Style nit: space after `for`

src/test/scala/com/databricks/spark/avro/AvroSuite.scala
```diff
@@ -443,4 +443,92 @@ class AvroSuite extends FunSuite {
     assert(newDf.count == 8)
   }
 
+  test("test doc in meta") {
+    val df = TestSQLContext.load(episodesFile, "com.databricks.spark.avro")
+    df.schema.fields(0).metadata.getString("_doc")
+
+    for(x <- df.schema.fields) {
```
JoshRosen: More style nits: always add a space after `for` and `if`.

marmbrus:
Thanks for working on this.
 - I'd consider making the addition of the alias columns just an option to the datasource, so that it is usable in languages other than scala.
 - Other users of metadata don't use `_` so i'd just go with `doc` and `aliases`

src/main/scala/com/databricks/spark/avro/SchemaConverters.scala
vasanthrj: used breakable for moving next element. primary checks before creating the fields. I feel this is simple than writing complex if condition check. I can modify if required.

vasanthrj:
@jendap Implemented review comments
Suggestion reply:
2. I added to dataframe since dataframe know the schema and can store metadata.Where aliases/doc are belong to a column data. I think parquet properties are related to deserialization, data type handling.

 @JoshRosen   Implemented review comments
@marmbrus adding alias column is optional where alias columns added after invoking addAvroAliasColumns()

Build is failing due to some other reason. Kindly check

src/main/scala/com/databricks/spark/avro/SchemaConverters.scala
```diff
@@ -29,6 +29,9 @@ import org.apache.avro.Schema.Type._
  */
 private object SchemaConverters {
 
+  val METADATA_KEY_DOC = "doc";
+  val METADATA_KEY_ALIASES = "aliases";
```
jendap: this is good, you can put even the "_parent" here as a constant

src/main/scala/com/databricks/spark/avro/SchemaConverters.scala
jendap: Yes, it is nice to check them before. But as far as simplicity and readability goes...
```scala
breakable {
  if (predicate) break
  body
}
```
is equivalent to
```scala
if (!predicate) {
  body
}
```
Hence is is shorter, cleaner and other folks do not need to know what breakable is. Right?

BTW: the code here can be written as...
```scala
val nonAliasStructFields = structType.fields.filterNot(field =>
  field.metadata.contains(METADATA_KEY_PARENT) && !field.metadata.getString(METADATA_KEY_PARENT).equals(field.name))
nonAliasStructFields.foreach { field =>
  var newFieldBuilder = fieldsAssembler.name(field.name)
...
```

src/test/scala/com/databricks/spark/avro/AvroSuite.scala
```diff
@@ -394,4 +394,93 @@ class AvroSuite extends FunSuite {
...
+      } else if (x.name == "air_date") {
+        assert("initial date" == x.metadata.getString(SchemaConverters.METADATA_KEY_DOC))
+      }
+    }
+  }
+
+  test("test aliases in meta") {
+    val df = TestSQLContext.load(testFile, "com.databricks.spark.avro")
+
+    for (x <- df.schema.fields) {
```
jendap: I guess people will believe the metadata are not type dependent. Hence you can put them only on one field in test file - let's say "string" and then this whole test can become
```scala
val df = TestSQLContext.load(testFile, "com.databricks.spark.avro")
assert(df.schema("string").getStringArray(SchemaConverters.METADATA_KEY_ALIASES) === Array("string_alias1", "string_alias1")
```
Would that make it easier?

vasanthrj:
@jendap Implemented your review comments. Please review.

marmbrus:
> @marmbrus adding alias column is optional where alias columns added after invoking addAvroAliasColumns()

I realize its optional, but implementing this feature as a scala function means that users in Python and R will not be able to use it.  I think instead this should just be a parameter that is passed in the `OPTIONS` clause or as an `option(...)` so that it works across languages.

vasanthrj:
@marmbrus Implemented review comments. Added a option. Please review

src/main/scala/com/databricks/spark/avro/package.scala
```diff
@@ -44,4 +45,12 @@ package object avro {
...
     def avro: String => DataFrame = reader.format("com.databricks.spark.avro").load
   }
+  
+    /**
+   * Adds a method, `avro`, to DataFrameReader that allows you to read avro files using
+   * the DataFileReade
+   */
+  implicit class AvroDataFrame(dataFrame: DataFrame) {
+    def addAvroAliasColumns(): DataFrame = SchemaConverters.dataFrameWithAliasColumn(dataFrame)
+  }
```
marmbrus: Can we get rid of this entirely?  I'd prefer to only have to support a single unified way to do this.

README.md
```diff
@@ -128,6 +126,18 @@ To save DataFrame as avro you should use the `save` method in `AvroSaver`. For e
...
+```scala
+scala>val dfWithAlias = df.addAvroAliasColumns()
+```
+In Spark SQL, loading avro with alias columns use:
+```
+create table mytable using com.databricks.spark.avro OPTIONS(path "test.avro", withAlias "true");
+```
+
```
marmbrus: We should also include examples of how to set alias and docs.

vasanthrj: @marmbrus user can set alias and docs to a avro using avro tools. Not using spark-avro

README.md
```diff
@@ -128,6 +126,19 @@ To save DataFrame as avro you should use the `save` method in `AvroSaver`. For e
 ```scala
 scala> AvroSaver.save(myRDD, "my/output/dir")
 ```
+
+Avro schema's field 'doc' will be preserved in DataFrame's schema metadata.
+
+To include aliases column in scheme invoke
```
marmbrus: To include extra columns when aliases are specified in the schema, set `withAlias` to `true`, as follows:

I would also call the property `withAliases`, since we handle more than one.

README.md
```diff
@@ -128,6 +126,19 @@ To save DataFrame as avro you should use the `save` method in `AvroSaver`. For e
 ```scala
 scala> AvroSaver.save(myRDD, "my/output/dir")
 ```
+
+Avro schema's field 'doc' will be preserved in DataFrame's schema metadata.
```
marmbrus: Make this its own section on extra metadata.

README.md
```diff
@@ -128,6 +126,19 @@ To save DataFrame as avro you should use the `save` method in `AvroSaver`. For e
...
+
+To include aliases column in scheme invoke
+```
+val df = sqlContext.read.format("com.databricks.spark.avro").option("withAlias", "true").load("test.avro")
+```
+In Spark SQL, loading avro with alias columns use:
```
marmbrus: This is probably not needed, but if you want to include it:

"The same can be done using `OPTIONS` passed to Spark SQL DDL:"

README.md
```diff
@@ -128,6 +126,19 @@ To save DataFrame as avro you should use the `save` method in `AvroSaver`. For e
...
+```
+create table mytable using com.databricks.spark.avro OPTIONS(path "test.avro", withAlias "true");
```
marmbrus: make the reserved words `CREATE` `TABLE`, etc capital to match the rest of the documentation.

README.md
```diff
@@ -128,6 +126,19 @@ To save DataFrame as avro you should use the `save` method in `AvroSaver`. For e
...
+```
+While saving, alias columns in DataFrame schema will be stored in `aliases` of avro file.
```
marmbrus: This is kind of confusing since we have our own concept of `alias` which you assign with `alias` or `as`.  We should probably have an example that shows setting metadata.

src/main/scala/com/databricks/spark/avro/AvroRelation.scala
```diff
@@ -68,7 +68,8 @@ private[avro] class AvroRelation(
    */
   override def dataSchema: StructType = maybeDataSchema match {
     case Some(structType) => structType
-    case None => SchemaConverters.toSqlType(avroSchema).dataType.asInstanceOf[StructType]
+    case None => SchemaConverters.toSqlType(avroSchema,
+        withAliasFields).dataType.asInstanceOf[StructType]
```
marmbrus: Style nit: don't break in the middle of a function call.  wrap the whole line instead.

src/main/scala/com/databricks/spark/avro/AvroRelation.scala
```diff
@@ -117,6 +118,7 @@ private[avro] class AvroRelation(
    * contain `requiredColumns`
    */
   override def buildScan(requiredColumns: Array[String], inputs: Array[FileStatus]): RDD[Row] = {
+    val withalias = withAliasFields
```
marmbrus: Is this just for serialization?  I'd probably just make sure to add transient elsewhere to match the style of other data sources.  Otherwise comment what is going on.

src/main/scala/com/databricks/spark/avro/AvroRelation.scala
```diff
@@ -135,7 +137,13 @@ private[avro] class AvroRelation(
               val superSchema = firstRecord.getSchema // the schema of the actual record
 
               // the fields that are actually required along with their converters
-              val avroFields = superSchema.getFields.map(f => (f.name, f)).toMap
+              var avroFields:Map[String, Field] = Map()
```
marmbrus: Style: space after `:`.

src/main/scala/com/databricks/spark/avro/SchemaConverters.scala
```diff
@@ -17,9 +17,7 @@ package com.databricks.spark.avro
 
 import java.nio.ByteBuffer
 import java.util.HashMap
-
```
marmbrus: Don't remove these.

src/main/scala/com/databricks/spark/avro/SchemaConverters.scala
```diff
@@ -27,6 +25,9 @@ import org.apache.avro.SchemaBuilder._
 import org.apache.avro.Schema.Type._
 import org.apache.spark.sql.Row
 import org.apache.spark.sql.types._
+import org.apache.spark.sql.DataFrame
+import java.util.ArrayList
```
marmbrus: sort these with other java and scala imports.

src/main/scala/com/databricks/spark/avro/SchemaConverters.scala
```diff
@@ -103,9 +136,22 @@ private object SchemaConverters {
       schemaBuilder: RecordBuilder[T],
       recordNamespace: String): T = {
     val fieldsAssembler: FieldAssembler[T] = schemaBuilder.fields()
-    structType.fields.foreach { field =>
-      val newField = fieldsAssembler.name(field.name).`type`()
 
+    val nonAliasStructFields = structType.fields.filterNot(field =>
+      field.metadata.contains(METADATA_KEY_ALIASES)
+        && field.metadata.contains(METADATA_KEY_PARENT)
+            && !field.metadata.getString(METADATA_KEY_PARENT).equals(field.name))
```
marmbrus: This indentation is inconsistent.

src/test/scala/com/databricks/spark/avro/AvroSuite.scala
```diff
@@ -3,9 +3,7 @@ package com.databricks.spark.avro
 import java.io.File
 import java.nio.ByteBuffer
 import java.sql.Timestamp
-
```
marmbrus: Don't remove these.

src/test/scala/com/databricks/spark/avro/AvroSuite.scala
```diff
@@ -16,6 +14,7 @@ import org.apache.spark.sql.test.TestSQLContext
 import org.apache.spark.sql.test.TestSQLContext._
 import org.apache.spark.sql.types._
 import org.scalatest.FunSuite
+import org.apache.spark.sql.AnalysisException
```
marmbrus: Sort these (and there is some existing lack of ordering).

src/test/scala/com/databricks/spark/avro/AvroSuite.scala
```diff
@@ -382,10 +403,49 @@ class AvroSuite extends FunSuite {
       assert(df.count == 8)
 
       val tempSaveDir = s"$tempDir/save/"
-
-      df.write.avro(tempSaveDir)
+            df.write.avro(tempSaveDir)
```
marmbrus: indentation is off

src/test/scala/com/databricks/spark/avro/AvroSuite.scala
```diff
@@ -382,10 +403,49 @@ class AvroSuite extends FunSuite {
...
+  test("test aliases in meta") {
+    val df = TestSQLContext.read.avro(testFile)
+    df.schema("string").metadata.getStringArray(SchemaConverters.METADATA_KEY_ALIASES) === Array("string_alias1", "string_alias1")
+  }
+
+  test("test aliases columns in data frame") {
+    var df = TestSQLContext.read.avro(testFile)
+    var fieldArray = df.schema.fieldNames;
+    assert(fieldArray contains("string"))
+    assert(!(fieldArray contains("string_alias1")))
```
marmbrus: Either use infix notation or leave the `.` in.

src/test/scala/com/databricks/spark/avro/AvroSuite.scala
```diff
@@ -382,10 +403,49 @@ class AvroSuite extends FunSuite {
...
+    assert(!(fieldArray contains("string_alias2")))
+    assert(!(fieldArray contains("map_alias")))
+    assert(!(fieldArray contains("enum_alias")))
+    assert(!(fieldArray contains("union_int_alias")))
+
+    fieldArray = SchemaConverters.dataFrameWithAliasColumn(df).schema.fieldNames
```
marmbrus: this test should probably use the user visible option instead of diving into internals.

src/test/scala/com/databricks/spark/avro/AvroSuite.scala
```diff
@@ -382,10 +403,49 @@ class AvroSuite extends FunSuite {
...
+  }
+
+  test("test aliases in meta") {
+    val df = TestSQLContext.read.avro(testFile)
+    df.schema("string").metadata.getStringArray(SchemaConverters.METADATA_KEY_ALIASES) === Array("string_alias1", "string_alias1")
+  }
+
+  test("test aliases columns in data frame") {
+    var df = TestSQLContext.read.avro(testFile)
+    var fieldArray = df.schema.fieldNames;
```
marmbrus: break these into two tests instead of using a `var`.

src/main/scala/com/databricks/spark/avro/AvroRelation.scala
```diff
@@ -135,7 +137,13 @@ private[avro] class AvroRelation(
...
+              superSchema.getFields.map(f => {
+                  avroFields += (f.name -> f)
+                  if (!f.aliases().isEmpty() && withalias) {
+                    f.aliases().foreach { alias => { avroFields += (alias -> f) } }
+                  }
+                })
```
marmbrus: Maybe something like?
```scala
val fields = superSchema.getFields.map(f => f.name -> f)
val aliasFields = if (withalias) {
  superSchema.getFields.flatMap(f => f.aliases.map(a => a -> f))
} else {
  Nil
}
val allFields = (fields ++ aliasFields).toMap
```

src/main/scala/com/databricks/spark/avro/SchemaConverters.scala
```diff
@@ -94,6 +114,19 @@ private object SchemaConverters {
     }
   }
 
+  def dataFrameWithAliasColumn(df : DataFrame) : DataFrame = {
```
marmbrus: Is this function still needed? or is this redundant with the above listing of field names ++ aliases?

marmbrus:
Thanks again for working on this!  I did another review pass.

src/main/scala/com/databricks/spark/avro/AvroRelation.scala
vasanthrj: while accessing directly class variable 'withAliasFields' inside the map 'superSchema.getFields.map' which calls hashcode() function. In spark 1.4.0, getting Class not found exception for guava class 'Objects' in hashcode() function.

marmbrus: That sounds like a problem with your executor classpath.

vasanthrj:
@marmbrus implemented review comments. Except the usage of "withAliasesFields" of in buildScan(...) of AvroRelation.scala, I was getting:
 object not serializable (class: org.apache.avro.Schema$RecordSchema,

gasparms:
What happened with this PR? Are you going to plan include this in future releases? It would be useful support aliases.

PR #61, opened 2015-07-13
Add a note about using sbt assembly when using --jars

tyro89:
This might be a useful note to add to the readme. In my case sbt assembly greatly simplified using spark-avro on an emr cluster. Maybe the note could help others? :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,10 @@ For example, to include it when starting the spark shell:
 $ bin/spark-shell --jars spark-avro_2.10-1.0.0.jar
 ```
 
+When using spark-avro through the `--jars` flag consider building spark-avro using sbt assembly. This will bundle all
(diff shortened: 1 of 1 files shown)


marmbrus:
As an alternative, should we just suggest using `--packages` instead?

tyro89:
@marmbrus :+1: that's much better. Would you say that `--packages` is the recommended way to link the jar vs `--jars`? I can update this section a little to mention both but put emphasis on `--packages`.

marmbrus:
yeah, i think since `packages` always handles dependencies I would prefer it to `jars`.

JoshRosen:
Given user confusion (see #64), we should update the docs to include `--packages` before 2.0.

JoshRosen:
The README was updated to mention `--packages` in 7c66855d55a00ae8b7f0ff5f329028f0b9ec2b3e, so I'm going to mark this as resolved.

PR #63, opened 2015-07-14
cross build 2.10 and 2.11

jendap:
Modified according to https://github.com/databricks/sbt-spark-package

It does publish both locally on "sbt +publishLocal". Hopefully sbt-spark-package will publish both to spark-packages.org.

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -1,5 +1,3 @@
-scalaVersion := "2.10.4"
-
 resolvers += Resolver.url("artifactory", url("http://scalasbt.artifactoryonline.com/scalasbt/sbt-plugin-releases"))(Resolver.ivyStylePatterns)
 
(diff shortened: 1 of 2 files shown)


project/plugins.sbt
```diff
@@ -8,7 +6,7 @@ resolvers += "sonatype-releases" at "https://oss.sonatype.org/content/repositori
 
 resolvers += "Spark Package Main Repo" at "https://dl.bintray.com/spark-packages/maven"
 
-addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.11.2")
+addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.12.0")
```
jendap: This is upgraded since it was complaining "There may be incompatibilities among your library dependencies."

marmbrus:
Thanks!

PR #66, opened 2015-07-19
Cleanup tests

jendap:
cleanup tests

* improve comparing avro records
* use FileUtils.deleteDirectory from a library
  (we do not need anything special here)


marmbrus:
Thanks!

PR #70, opened 2015-08-06
Added Partitioning to work with Avro

JDrit:
This library currently does not work with partitioning, it will fail silently due to what I expect is multiple writers trying to write to the same file. 

This PR modifies the record writer to use unique file names so this does not happen.

### DIFF
--- a/src/main/scala/com/databricks/spark/avro/AvroRelation.scala
+++ b/src/main/scala/com/databricks/spark/avro/AvroRelation.scala
@@ -17,7 +17,10 @@
 package com.databricks.spark.avro
 
 import java.io.FileNotFoundException
+import java.util.UUID
 import java.util.zip.Deflater
+
+import org.apache.hadoop.mapreduce.lib.output.FileOutputFormat
 
 import scala.collection.Iterator
 import scala.collection.JavaConversions._
@@ -49,6 +52,13 @@ private[avro] class AvroRelation(
     override val userDefinedPartitionColumns: Option[StructType],
     private val parameters: Map[String, String])
     (@transient val sqlContext: SQLContext) extends HadoopFsRelation with Logging {
+
+  // This UUID is used to avoid output file name collision between different appending write jobs.
+  // These jobs may belong to different SparkContext instances. Concrete data source implementations
+  // may use this UUID to generate unique file names (e.g., `part-r-<task-id>-<job-uuid>.parquet`).
+  //  The reason why this ID is used to identify a job rather than a single task output file is
+  // that, speculative tasks must generate the same output file name as the original task.
+  private val uniqueWriteJobId = UUID.randomUUID().toString
 
   private val IgnoreFilesWithoutExtensionProperty = "avro.mapred.ignore.inputs.without.extension"
   private val recordName = parameters.getOrElse("recordName", "topLevelRecord")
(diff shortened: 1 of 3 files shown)


marmbrus:
Seems reasonable to me, but I'd like @liancheng to look at it as he did the uuid stuff in Spark.

JDrit:
@marmbrus @liancheng just wanted to see if you have any feedback on this.

src/main/scala/com/databricks/spark/avro/AvroRelation.scala
```diff
@@ -50,6 +53,13 @@ private[avro] class AvroRelation(
     private val parameters: Map[String, String])
     (@transient val sqlContext: SQLContext) extends HadoopFsRelation with Logging {
 
+  // This UUID is used to avoid output file name collision between different appending write jobs.
+  // These jobs may belong to different SparkContext instances. Concrete data source implementations
+  // may use this UUID to generate unique file names (e.g., `part-r-<task-id>-<job-uuid>.parquet`).
+  //  The reason why this ID is used to identify a job rather than a single task output file is
+  // that, speculative tasks must generate the same output file name as the original task.
+  private val uniqueWriteJobId = UUID.randomUUID().toString
```
liancheng: The unique write job ID is handled within `BaseWriterContainer`, and every job writes to a `HadoopFsRelation` should have a different UUID. Is there any other reason that we have this UUID within `AvroRelation`?

src/test/scala/com/databricks/spark/avro/AvroSuite.scala
```diff
@@ -21,6 +22,20 @@ class AvroSuite extends FunSuite {
   val episodesFile = "src/test/resources/episodes.avro"
   val testFile = "src/test/resources/test.avro"
 
+  test("reading and writing partitioned data") {
+    TestUtils.withTempDir { dir =>
```
liancheng: How about moving the `withTempDir` call into the `for` loop? So that you don't need the UUID trick to create unique directories.

liancheng:
Sorry for my late reply, only have a question about the position of the write job UUID.

src/main/scala/com/databricks/spark/avro/AvroRelation.scala
JDrit: The reason was that it seemed that the `BaseWriterContainer` was not setting the UUID so I was setting it myself.

liancheng: Hm, that's weird. The UUID is set [here] [1]. Could you please provide a snippet that reproduces the issue you mentioned? I'll also try to reproduce it locally.

On the other hand, adding the UUID here isn't appropriate, because we'd like to use the UUID to identify each write job rather than each relation. We don't want any two write job have the same UUID.

[1]: https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/WriterContainer.scala#L88-L91

liancheng: @JDrit The reason why you can't retrieve the UUID is that, the Spark version defined in `build.sbt` is 1.4.0, while the UUID was added in 1.4.1 to fix a severe bug that may cause data corruption.

liancheng:
@JDrit I sent out #77 to bump Spark version to 1.4.1 and fixed the write job UUID.

JDrit:
Got it, that works. Should I closed this PR then?

liancheng:
Yeah, I think we can close this PR now. Thanks for working on this!

PR #72, opened 2015-08-13
Docs

JDrit:
I updated the docs to match the new changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,18 +7,33 @@ A library for querying Avro data with [Spark SQL](http://spark.apache.org/docs/l
 
 ## Requirements
 
-This library requires Spark 1.3+. There is a 1.2 version as well.
+This documentation is for Spark 1.4+.
+
+This library has different versions for 1.2, 1.3, and 1.4.
+
+### Versions
+Spark changed how it reads / writes data in 1.4, so please use the correct version
+of this dedicated for your spark version
+
+1.2  -> `0.2.0`
+
+1.3 -> `1.0.0`
+
+1.4+ -> `1.1.0`
 
 ## Linking
 You can link against this library (for Spark 1.3+) in your program at the following coordinates:
 
-```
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -85,58 +114,89 @@ TimestampType -> long
...
+You can specify the record name and namespace like this:
+
 ```scala
-scala> AvroSaver.save(myRDD, "my/output/dir", Map("recordName" -> "MyRecord", "recordNamespace" -> "com.mycompany.mystuff"))
-```
+import com.databricks.spark.avro._
 
-We also support the ability to read all avro files from some directory. To do that, you can pass
-a path to that directory to the avroFile() function. However, there is a limitation - all of
-those files must have the same schema. Additionally, files used must have a .avro extension.
```
nlande: Question: Is "all of those files must have the same schema." no longer a limitation?

rdblue: It should be possible to build an overall merged schema, but it is better to have an idea of the overall schema like you would get with Hive (table DDL) or Kite (which explicitly stores an overall Avro schema). If you need to generate a merged schema,  see [`SchemaUtil.merge`](https://github.com/kite-sdk/kite/blob/master/kite-data/kite-data-core/src/main/java/org/kitesdk/data/spi/SchemaUtil.java#L452). That method will merge schemas if they appear related (share compatible fields) and throw an exception if they don't.

nlande: Thanks rdblue, I will check out Kite. The usecase for us is that our avro schema evolves over time as we add or retire features. So some of our partitions will be stored with an older schema but if the newest schema is used to load them everything works fine with either Hive or base Spark. 

I might be insufficiently familiar with this library but since it appears as if there is no option to explicitly specify the load schema, there is no good way to load data across different partitions with compatible but different schemas.

marmbrus:
Thanks!  Merging to master.

Issue #54, opened 2015-06-23
Save DF with nested records with the same name


sixers:
spark-avro fails to save this dataset:

```python
df = sqlc.createDataFrame([Row(id=1, data=Row(data=Row(key='value')))])
df.save("s3://...", "com.databricks.spark.avro")
```
with the following error:

```
Py4JJavaError: An error occurred while calling o1873.save.
: org.apache.avro.SchemaParseException: Can't redefine: data
	at org.apache.avro.SchemaBuilder$NameContext.put(SchemaBuilder.java:936)
	at org.apache.avro.SchemaBuilder$NameContext.access$600(SchemaBuilder.java:884)
	at org.apache.avro.SchemaBuilder$NamespacedBuilder.completeSchema(SchemaBuilder.java:470)
	at org.apache.avro.SchemaBuilder$RecordBuilder.fields(SchemaBuilder.java:1734)
	at com.databricks.spark.avro.SchemaConverters$.convertStructToAvro(SchemaConverters.scala:99)
	at com.databricks.spark.avro.SchemaConverters$.com$databricks$spark$avro$SchemaConverters$$convertFieldTypeToAvro(SchemaConverters.scala:190)
	at com.databricks.spark.avro.SchemaConverters$$anonfun$convertStructToAvro$1.apply(SchemaConverters.scala:104)
	at com.databricks.spark.avro.SchemaConverters$$anonfun$convertStructToAvro$1.apply(SchemaConverters.scala:100)
	at scala.collection.IndexedSeqOptimized$class.foreach(IndexedSeqOptimized.scala:33)
	at scala.collection.mutable.ArrayOps$ofRef.foreach(ArrayOps.scala:108)
	at com.databricks.spark.avro.SchemaConverters$.convertStructToAvro(SchemaConverters.scala:100)
	at com.databricks.spark.avro.SchemaConverters$.com$databricks$spark$avro$SchemaConverters$$convertFieldTypeToAvro(SchemaConverters.scala:190)
	at com.databricks.spark.avro.SchemaConverters$$anonfun$convertStructToAvro$1.apply(SchemaConverters.scala:104)
	at com.databricks.spark.avro.SchemaConverters$$anonfun$convertStructToAvro$1.apply(SchemaConverters.scala:100)
	at scala.collection.IndexedSeqOptimized$class.foreach(IndexedSeqOptimized.scala:33)
	at scala.collection.mutable.ArrayOps$ofRef.foreach(ArrayOps.scala:108)
[...]

PR #73, opened 2015-08-14
Added differentiation of nested records with the same name

reyven:
by adding to each record definition a namespace definition consisting of a
dot-separated concatenation of the enclosing records' names.

This fixes https://github.com/databricks/spark-avro/issues/54

The implementation adheres to the Avro name/namespace specification
at https://avro.apache.org/docs/1.7.7/spec.html#Names
and uses the fullname definition (as opposed to seperate name and
namespace definition) to preserve union type resolution.

### DIFF
--- a/src/main/scala/com/databricks/spark/avro/SchemaConverters.scala
+++ b/src/main/scala/com/databricks/spark/avro/SchemaConverters.scala
@@ -235,8 +235,8 @@ private object SchemaConverters {
       case structType: StructType =>
         convertStructToAvro(
           structType,
-          schemaBuilder.record(recordNamespace+"."+structName),
-          recordNamespace+"."+structName)
+          schemaBuilder.record(recordNamespace + "." + structName),
+          recordNamespace + "." + structName)
 
       case other => throw new IllegalArgumentException(s"Unexpected type $dataType.")
     }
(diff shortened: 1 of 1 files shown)


JoshRosen:
This failed code style checks; please see the Travis output.

JoshRosen:
Can you add a regression test for this bug?

vlyubin:
Ping @reyven - it would be nice to merge this once we have a test.

H4ml3t:
Any update? I'm stuck with this issues.

dmitriy-svds:
Hello all,

It would be great to have this merged. I'm having the same issue.

Thanks!

JoshRosen:
There's no way that we can safely merge this without tests. It would be great if someone could take over this PR to contribute tests for this feature.

vlyubin:
Here's a simple test that can be added to AvroSuite:
```scala
  // Classes have to be defined outside of the test
  case class NestedBottom(id: Int, data: String)
  case class NestedMiddle(id: Int, data: NestedBottom)
  case class NestedTop(id: Int, data: NestedMiddle)

  test("saving avro that has nested records with the same name") {
    TestUtils.withTempDir { tempDir =>
      val df = TestSQLContext.createDataFrame(List(NestedTop(1, NestedMiddle(2, NestedBottom(3, "1")))))
      df.write.avro(s"$tempDir/duplicate_names/")
    }
  }
```

Unfortunately, it seems like the current solution doesn't resolve the issue, the test above fails with `UnresolvedUnionException` :(

dotcomputercraft:
Hello all,  Can we merge this PR?

lyubinetz:
@dotcomputercraft I don't think this really works.

https://github.com/databricks/spark-avro/pull/249 was just proposed and on the first glance it actually works!

gengliangwang:
I think this issue will be resolved in #249 .
Close this PR since it doesn't work and there is no update.

PR #74, opened 2015-08-19
fix read multiple avro files at the same time

yucheng1992:
Hi all, I find that the avroFile function in package.scala doesn't work although it has a deprecated annotation. If this is not the right place let me know and I will move it. Right here, I just fix avroFile function to let it read multiple Avro files at the same time. You can use either a Sequence of file names or file names separated by comma.


JoshRosen:
I'm going to resolve this as "won't fix" because this functionality is deprecated in 2.x versions of this library and is removed in 3.x versions. Sorry!

PR #78, opened 2015-08-27
Minimize binary exception surface

marmbrus:
In order to ease binary compatibility down the road, I've replaced a bunch of Avro specific exceptions with appropriate java equivalents.  In the process I changed some pretty dense functional code that was running in the critical path, to a more explicit imperative version.

### DIFF
--- a/src/main/scala/com/databricks/spark/avro/AvroRelation.scala
+++ b/src/main/scala/com/databricks/spark/avro/AvroRelation.scala
@@ -38,12 +38,6 @@ import org.apache.spark.sql.sources._
 import org.apache.spark.sql.types.StructType
 import org.apache.spark.sql.{Row, SQLContext}
 
-abstract class AvroRelationException(msg: String) extends Exception(msg)
-case object NoFilesException extends AvroRelationException("no input files given")
-case class SchemaConversionException(msg: String) extends AvroRelationException(msg)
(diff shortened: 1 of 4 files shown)


src/main/scala/com/databricks/spark/avro/AvroRelation.scala
```diff
@@ -134,20 +129,54 @@ private[avro] class AvroRelation(
...
+                    val field =
+                      avroFieldMap.getOrElse(
+                        columnName,
+                        throw new AssertionError(s"Invalid column $columnName"))
+                    val converter = SchemaConverters.createConverterToSQL(field.schema)
+
+                    (record: GenericRecord) => rowBuffer(idx) = converter(record.get(field.pos()))
+                }
+
+                private def advanceNextRecord() = {
```
JoshRosen: Just curious, but why did you choose this `advanceNext()` idiom over `def hasNext = baseIterator.hasNext` etc?

JoshRosen:
LGTM overall.

src/main/scala/com/databricks/spark/avro/AvroRelation.scala
marmbrus: This was so that we could still feed in the row that we took off the base iterator earlier (in order to know the schema of this particular split).

Issue #19, opened 2015-01-30
Include avro-mapred-1.7.7-hadoop1.jar in dependencies


gy8:
For us noobs who are just starting out, it would be really helpful to explicitly state in the readme that in order to run the example stated you need to download and include avro-mapred-1.7.7-hadoop1.jar in addition to spark-avro.

Apologies if this was a really noob mistake but I figured this might save others some confusion in the future.

Awesome tool and keep up the good work!

gy8:
Thanks Volodymyr for getting back to me! I think this was mostly likely due to having installed spark via cloudera v5. Sorry for being a noob

PR #79, opened 2015-08-28
Allow a single published artifact to work with multiple Hadoop versions

JoshRosen:
This commit allows us to publish one `spark-avro` artifact which is built against a fixed Hadoop version but which works with both Hadoop 1.x and 2.x. In the past, we published separate artifacts for Hadoop 1.x and 2.x in order to work around a binary incompatibility in `TaskAttemptContext` (see #19). This patch works around the incompatibility using reflection, similar to https://github.com/apache/spark/pull/6599.

In order to make this testable, this patch also modifies our SBT build and Travis configuration so that the test Hadoop dependencies can be configured separately from the compile dependency.

I made a similar fix in `spark-redshift`: https://github.com/databricks/spark-redshift/pull/58

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,26 +11,26 @@ matrix:
     # Spark 1.4.1 and Scala 2.10
     - jdk: openjdk6
       scala: 2.10.4
-      env: HADOOP_VERSION="2.2.0" TEST_SPARK_VERSION="1.4.1"
+      env: TEST_HADOOP_VERSION="2.2.0" TEST_SPARK_VERSION="1.4.1"
     - jdk: openjdk7
       scala: 2.10.4
-      env: HADOOP_VERSION="2.2.0" TEST_SPARK_VERSION="1.4.1"
+      env: TEST_HADOOP_VERSION="2.2.0" TEST_SPARK_VERSION="1.4.1"
     # Spark 1.5.0 and Scala 2.10
     - jdk: openjdk7
       scala: 2.10.4
-      env: HADOOP_VERSION="1.0.4" SPARK_VERSION="1.5.0-rc2"
+      env: TEST_HADOOP_VERSION="1.0.4" TEST_SPARK_VERSION="1.5.0-rc2"
     - jdk: openjdk7
       scala: 2.10.4
-      env: HADOOP_VERSION="1.2.1" SPARK_VERSION="1.5.0-rc2"
+      env: TEST_HADOOP_VERSION="1.2.1" TEST_SPARK_VERSION="1.5.0-rc2"
     - jdk: openjdk7
       scala: 2.10.4
-      env: HADOOP_VERSION="2.2.0" SPARK_VERSION="1.5.0-rc2"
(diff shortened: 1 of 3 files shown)


JoshRosen:
(This initial commit will intentionally fail tests to demonstrate that the Hadoop version switching in the matrix builds is configured properly.)

JoshRosen:
I added reflection hacks in the `spark-avro` code but the tests are still going to fail, even under 1.5.0, because it looks like BaseWriterContainer still calls one of the TaskAttemptContext methods without using reflection: https://github.com/apache/spark/blob/1f90c5e2198bcf49e115d97ec300c17c1be4dcb4/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/WriterContainer.scala#L230

If we cut another 1.5.x release candidate then I'd like to fix this.

/cc @marmbrus @liancheng

JoshRosen:
It looks like this is going to be harder than I thought (maybe impossible?) because individual `avro-mapred` binaries are not compatible with both Hadoop 1.x and 2.x

`spark-redshift` worked around this problem by declaring a `provided` dependency on `avro-mapred`.

`org.apache.spark.spark-hive` seems to depend on `avro-mapred`, though, so things may happen to work out okay for users if they run against a Spark version that has been compiled against the Hadoop version that they're using.

Given that `spark-redshift`'s own incompatibilities can be worked around using reflection, if we ended up publishing multiple releases I think that the only difference would be in how our pom declares the `avro-mapred` dependency.

JoshRosen:
It looks like Avro's Hadoop version incompatibilities also stem from the changes in TaskAttemptContext. It appears that they also tried to work around this using reflection but this doesn't appear to have properly fixed the problem.  See https://issues.apache.org/jira/browse/AVRO-1170 and https://issues.apache.org/jira/browse/AVRO-1230.

I'd actually like to see this eventually get worked around upstream in avro-mapred, since the current status quo seems to cause a lot of problems for many users; see https://issues.apache.org/jira/browse/AVRO-1506, http://stackoverflow.com/questions/29448222/found-interface-org-apache-hadoop-mapreduce-taskattemptcontext, http://stackoverflow.com/questions/30115108/amazon-hadoop-2-4-avro-1-77-found-interface-org-apache-hadoop-mapreduce-task, and http://stackoverflow.com/questions/19089557/how-does-cloudera-cdh4-work-with-avro for a few examples (there are literally pages of examples if you search for "avro taskattemptcontext" on Google).

JoshRosen:
Fortunately, there might be a simple fix for us: just switch `avro-mapred`'s dependency back to `provided`.  It used to be marked as provided in 1.0.0, but it looks like this was changed (unintentionally?) in a recent patch: https://github.com/databricks/spark-avro/pull/59/files#r38227562

I think that 1.0.0 worked correctly because `spark-hive` happens to depend on `avro-mapred`, so at runtime the Spark assembly JAR plus the Hadoop classpath will provide the correct version of this dependency.

PR #81, opened 2015-08-29
Quick and dirty port of the merge script

marmbrus:
@JoshRosen if you have a more cleaned up version of this we could commit that instead, but this at least has been working for me.

### DIFF
--- a/dev/merge_pr.py
+++ b/dev/merge_pr.py
@@ -1,0 +1,451 @@
+#!/usr/bin/env python
+
+#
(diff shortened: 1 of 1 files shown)


JoshRosen:
Seems fine to me; I suppose we could remove the JIRA integration stuff, etc, but not a high priority yet since we're the only ones using it for now.

marmbrus:
Yeah, I'm inclined to merge and remove more later since having untracked stuff breaks the new release process.

JoshRosen:
Yep, let's merge.

PR #82, opened 2015-08-29
Update SBT lib to newest version in Spark

marmbrus:
Moves the sbt scripts to `build/` to match Spark and update it to reflect the new download locations.


build/mvn
```diff
@@ -0,0 +1,152 @@
+#!/usr/bin/env bash
```
JoshRosen: I don't think we need build/mvn.

JoshRosen:
LGTM pending Travis.

PR #83, opened 2015-08-29
Add command for publishing a release

marmbrus:
Add the sbt release command and configure it to sign and also publish to Spark Packages.  In the process I cleaned up some stale plugins that were causing library conflicts.

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -20,6 +20,8 @@ addSbtPlugin("org.spark-packages" % "sbt-spark-package" % "0.2.2")
 
 addSbtPlugin("org.scoverage" % "sbt-scoverage" % "1.1.0")
 
+addSbtPlugin("org.scalastyle" %% "scalastyle-sbt-plugin" % "0.6.0")
+
 addSbtPlugin("me.lessis" % "bintray-sbt" % "0.3.0")
 
 addSbtPlugin("com.github.gseitz" % "sbt-release" % "1.0.0")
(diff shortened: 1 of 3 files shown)


project/plugins.sbt
```diff
@@ -6,35 +6,24 @@ resolvers += "sonatype-releases" at "https://oss.sonatype.org/content/repositori
 
 resolvers += "Spark Package Main Repo" at "https://dl.bintray.com/spark-packages/maven"
 
-addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.12.0")
+libraryDependencies += "org.ow2.asm"  % "asm" % "5.0.3"
```
JoshRosen: Do we still need these?

project/plugins.sbt
```diff
@@ -6,35 +6,24 @@ resolvers += "sonatype-releases" at "https://oss.sonatype.org/content/repositori
...
 addSbtPlugin("com.typesafe.sbteclipse" % "sbteclipse-plugin" % "4.0.0")
 
 addSbtPlugin("com.github.mpeltonen" % "sbt-idea" % "1.6.0")
 
-// For Sonatype publishing
-//resolvers += Resolver.url("sbt-plugin-releases", new URL("http://scalasbt.artifactoryonline.com/scalasbt/sbt-plugin-releases/"))(Resolver.ivyStylePatterns)
+addSbtPlugin("com.jsuereth" % "sbt-pgp" % "1.0.0")
 
-//addSbtPlugin("com.jsuereth" % "xsbt-gpg-plugin" % "0.6")
+addSbtPlugin("org.spark-packages" % "sbt-spark-package" % "0.2.2")
```
JoshRosen: version 0.2.3 is out now, I think, but I don't think it has any changes relevant to this library (only R stuff).

marmbrus: updated

project/plugins.sbt
marmbrus: removed and it seems to work still

Issue #88, opened 2015-09-11
Expose the GenericRecord to Row conversion


fvlankvelt:
The AvroRelation class contains some very useful code to convert data from a GenericRecord to a Row representation.  This functionality is purely about Spark SQL / Avro interoperability.  However, it is currently tied to hdfs (the buildScan method in AvroRelation).  It would be most convenient if this code were extracted so that it could also be used with other persistence engines.

Concretely, I'm attempting to read Avro records from Kafka.

JoshRosen:
This seems related to #34?

fvlankvelt:
That issue seems to more about writing avro files with a specific schema.  This one is about reading avro records from a non-hdfs storage engine.

The issues are related in that both wish to leverage the spark-avro code by means of a clean api.

I'll try to come up with a PR.

JoshRosen:
Per my comment at https://github.com/databricks/spark-avro/pull/89#issuecomment-263147247, I think the right approach here is to expose an AvroSchema to `ExpressionEncoder[GenericRow]` interface, as described in #169. Thus, I'm going to close this as "won't fix" but will gladly accept help on #169.

PR #89, opened 2015-09-12
Expose the GenericRecord to Row conversion - #88

fvlankvelt:
This transformation makes it possible to use alternative storage engines (backing the RDD[GenericRecord]) for an Avro backed DataFrame.

### DIFF
--- a/src/main/scala/com/databricks/spark/avro/package.scala
+++ b/src/main/scala/com/databricks/spark/avro/package.scala
@@ -50,7 +50,8 @@ package object avro {
   }
 
   /**
-   * Adds a signature to [[SQLContext.createDataFrame]], to create a DataFrame from an RDD of GenericRecords.
+   * Adds a signature to [[SQLContext.createDataFrame]], to create a DataFrame from an RDD of
+   * GenericRecords.
    */
   implicit class AvroSQLContext(sqlContext: SQLContext) {
     def createDataFrame(rdd: RDD[GenericRecord], schema: Schema): DataFrame = {
--- a/src/main/scala/com/databricks/spark/avro/package.scala
+++ b/src/main/scala/com/databricks/spark/avro/package.scala
@@ -15,7 +15,12 @@
  */
 package com.databricks.spark
 
-import org.apache.spark.sql.{SQLContext, DataFrameReader, DataFrameWriter, DataFrame}
+import org.apache.avro.generic.GenericRecord
+import org.apache.spark.rdd.RDD
+import org.apache.spark.sql._
+import scala.collection.JavaConversions._
+
+import scala.collection.Iterator
 
 package object avro {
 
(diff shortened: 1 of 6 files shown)


src/main/scala/com/databricks/spark/avro/package.scala
```diff
@@ -15,7 +15,12 @@
  */
 package com.databricks.spark
 
-import org.apache.spark.sql.{SQLContext, DataFrameReader, DataFrameWriter, DataFrame}
+import org.apache.avro.generic.GenericRecord
+import org.apache.spark.rdd.RDD
+import org.apache.spark.sql._
+import scala.collection.JavaConversions._
```
JoshRosen: Nit: import ordering. Please group the Scala imports in their own section prior to the other imports.

JoshRosen:
@marmbrus, is this an API that we want to commit to making stable?

src/main/scala/com/databricks/spark/avro/package.scala
```diff
@@ -44,4 +49,68 @@ package object avro {
   implicit class AvroDataFrameReader(reader: DataFrameReader) {
     def avro: String => DataFrame = reader.format("com.databricks.spark.avro").load
   }
+
+  /**
+   * Adds a method, `toRowRDD`, to RDDs of GenericRecords.  This allows you to convert it to
+   * an RDD of Row (and, by implication, a DataFrame).
+   */
+  implicit class GenericRecordRDD(rdd: RDD[GenericRecord]) {
+    def toRowRDD(requiredColumns: Array[String]): RDD[Row] = rdd.mapPartitions { records =>
```
marmbrus: Could we instead make this `toDF(): DataFrame`?  I think that gives us more long term flexibility, for example, if we wanted to skip the public row representation and go directly to the internal row representation.  It also keeps the user from manually having to handle the schema when doing the conversion.

marmbrus: As long as we are going to expose this, we should also consider implementing it as a static method in an object (we can still have a convenience implicit) so that its usable from Java.

fvlankvelt: creating a DataFrame rather than an RDD[Row] certainly works for me.

Hadn't thought about the Java compatibility - thanks!

fvlankvelt:
I've adjusted the PR to account for the comments.  Instead of converting to an RDD[Row], the API now converts to a DataFrame.  It's also been made accessible from Java.

FRosner:
@JoshRosen @fvlankvelt  what is the status here?

flo076:
Hello @JoshRosen, have you any idea when this pull request will be merged ? and release ? 
Thanks you.

JoshRosen:
Not sure; I'm not actively reviewing for this project right now. You'll have to ask @marmbrus or @vlyubin regarding timetables / prioritization.

flo076:
@JoshRosen Thanks you for your answer. 
So @marmbrus / @vlyubin can you watch your timetables / prioritization and give me any informations please ? ( sorry for my english ;) )

Thanks you.

marmbrus:
I'm hesitant to expose internal details like this without thinking more about the API, especially when new APIs like Datasets/Encoders might make it much easier to work with Avro in a more principled way.  If you really need this logic, I'd consider just inlining it in your code so you aren't sensitive to changes we might make to the internals of this library (i.e. we might start converting directly to Tungstens format using an Encoder).

FRosner:
Ok works for me @marmbrus . So I don't have to wait for it and will copy and paste the code.

Thanks!

Dinduks:
@fvlankvelt Thanks for the great work!

Just another +1 for this PR. Hope it will be merged!

vinothchandar:
Any updates on this ? we have avro records in a kafka topic, that we would like to SQL on.. it would be great for this..

marmbrus:
I think for the kafka use case, it would make more sense to implement this like the `to_json` and `from_json` functions that were just added in Spark 2.1.  If you are looking for Specific or Generic records, then I think we should add encoders that can construct those objects.  Basically, my goal is to make sure that the interfaces we expose here are similar to the ones in core Spark, rather than something one-off to this library.

Thoughts?  Would that solve peoples use cases?

JoshRosen:
So I've given this a second look and I don't think that this is necessarily an unreasonable thing to expose given that data sources like CSV and JSON have methods for reading from `RDD[String]`, etc. I'm going to explore opening up a narrow interface for this in my own PR, but first I'll comment here on a couple of API considerations that I want to address in my PR.

JoshRosen:
In terms of API layering, it seems like all that we really need is a mechanism to take a user-provided Avro schema and produce an encoder which operates over objects that subclass `GenericRecord`. I'll see if there's a clean way to accomplish this layering.

JoshRosen:
After looking into this some more, I believe that porting the read path to use ExpressionEncoder is probably the right long-term approach here. At https://github.com/databricks/spark-avro/issues/169#issuecomment-253343393, @bdrillard has more-or-less done this (see https://github.com/databricks/spark-avro/compare/master...bdrillard:avro_encoder) and I think that we should aim to either port their change upstream or produce our own similar patch.

This will likely have performance improvements as well, since an expression-based encoder would be able to benefit from whole-stage codegen.

Given all of this, I'm going to close this PR as "won't fix" but will gladly accept PRs which advance the encoder-based approach. I was hoping that I'd be able to pull this in for a 3.1.0 release but it looks like this is going to take longer than I expected and I don't want to further delay an official release of master's outstanding changes. However, once we do merge an encoder-based approach then I can cut a new 3.2.0 release right away.

PR #97, opened 2015-11-03
Typo in maven dependency example

jiekebo:
Was wondering why maven wouldn't resolve the dependency :)


JoshRosen:
Good catch! Thanks for fixing this.

Issue #96, opened 2015-10-30
Specifying a read schema with spark-avro


mwho:
It would be nice to have an option to supply a read schema (in lieu of the embedded schema) when reading avro files via spark-avro.

For example, the Python Avro API allows the following:
`reader = DataFileReader(data, DatumReader(readers_schema=schema))`

The scenario is this: I have many .avro files, possibly with different schemas (due to schema evolution), and I would like to use a single "master" schema to ingest all of those avro files into a single Spark Dataframe.

theouteredge:
This will be a really good idea.
I've just been looking at Avro schema evolution and how I can manage this in spark

jongyonkim:
I'm waiting for this feature as well. Otherwise there is no way of ingesting many avro files (each with their own schemas) with an up-to-date master schema.

cleaton:
i'm also waiting for this feature as it seems spark is very slow in generating dataframe schema when a large number of sequence files are selected. It also seems to create thousands of broadcast variables and using a lot of memory on the driver node.

tomseddon:
Hi, is there any progress on this feature?  The only reason I'm not using spark-avro (using hadoopRDD instead) is because I need support for schemas that can evolve.  I'm sometimes re-processing historical data which may not have some attributes which have been recently added, but the whole lot would be processed in one go ideally.

clockfly:
In Spark 2.0, you can specify a schema using `spark.read.schema(user_defined_schema).format(...).`

```
scala> val df = Seq((1,2,3)).toDF("a", "b", "c")
df: org.apache.spark.sql.DataFrame = [a: int, b: int ... 1 more field]

scala> df.write.format("com.databricks.spark.avro").save("/tmp/output2")

scala> spark.read.format("com.databricks.spark.avro").load("/tmp/output2").show()
+---+---+---+
|  a|  b|  c|
+---+---+---+
|  1|  2|  3|
+---+---+---+

scala> import org.apache.spark.sql.types._
import org.apache.spark.sql.types._

// Prepare a customized schema
scala> val struct = StructType(StructField("a", IntegerType, true) :: Nil)
struct: org.apache.spark.sql.types.StructType = StructType(StructField(a,IntegerType,true))

scala> spark.read.format("com.databricks.spark.avro").schema(struct).load("/tmp/output").show()
+---+
|  a|
+---+
|  1|
+---+

// Save the schema to a Json string, you can later save this to a file.
scala> struct.json
[...]

Issue #52, opened 2015-06-07
Save DF with specific Avro schema


sixers:
Currently spark-avro constructs avro schema from Spark schema on save. It's limited, e.g. we can't set namespace and name of nested records. 

One scenario is reading Avro DF, and then saving it again: original schema, and new schema are different.

Having an option to pass Avro schema explictly should solve this.

rxin:
I'm not super familiar with avro - can avro schema be defined as a string? If yes, maybe we can just take a string option.

@sixers would you like to give this a try?

sixers:
@rxin Avro schema is just a json.

I'm not super familiar with Scala, but I can give it a try.

lukdumi:
Voting and waiting for this :+1:

mnarrell:
Any progress on this issue?

cbyn:
Would also love to see this

JoshRosen:
This turns out to be pretty easy to support now that we we honor user-provided Avro schemas during writes. I've opened #197 to fix this for 3.1.0.

JoshRosen:
After looking at this some more, I'm not entirely clear on what the semantics should be for a few important corner cases where there is significant mismatch between the Spark schema and user-provided schema (see https://github.com/databricks/spark-avro/pull/197#issuecomment-263152929). It would be great if folks who are interested in this feature could provide references which explain the desired behavior for the cases described in my linked comment.

Give that I think that this will need more design so that we don't build the wrong thing, I'm going to let this slip to 3.2.0.

LazyBoy248:
Hi Josh,

This is a feature that I'm looking for.  I'd like to associate a schema on write to enforce that the data reflects the schema.  If the data does not conform to the schema, I'd like spark to throw an exception since that's an indication that the transformation of my spark job is not aligned the expected data.

For example, let's say my spark data contain the following data
{
A: Football,
B: Soccer
}
If My schema was simply 

A:

I'd like my output to be 

{
A: Football
}

If A was missing, I'd like an exception so that I know my transformation is not creating A in my object.  I was wondering what's the hold up on this feature.

snadorp:
Have to agree with @LazyBoy248  here, the user supplied schema should rule over the generated Spark schema.

lindblombr:
[...]

PR #109, opened 2015-12-26
ISSUE #96: Specifying a read schema

yanxiaole:
Addressing issue #96 .

My solution is, err, a bit violent, all unknown columns were set to `null` value instead throw a exception, hope some one can improve it. Right now it works for me. Here's the test example:

```scala
import org.apache.avro.Schema

val avroSchema = new Schema.Parser().parse(new File("avsc/user.avsc"))
val df = sqlContext.read.schema(avroSchema).avro("avro/")
```

PR #106, opened 2015-12-01
ISSUE #105: support unions of a single type

dbieber:
#105 
Adds support in SchemaConverters for unions of a single type.

PR #113, opened 2016-01-24
Add a new feature to get Schema from explicit Avro class

ogirardot:
Hi, 
this PR adds the possibility to use an existing Avro schema from the corresponding generated Java class + it adds a new implicit to create a Dataframe from an existing RDD of Avro records.

It should fix : https://github.com/databricks/spark-avro/issues/96 and https://github.com/databricks/spark-avro/issues/52
Regards, 

Olivier.


ogirardot:
build errors are linked to scalariform versions c.f. https://github.com/sbt/sbt/issues/2089

JoshRosen:
I can fix the style issues by bumping the Scalastyle version in this repo. Note, however, that there are still a number of legitimate style problems in this PR; please see https://github.com/databricks/scala-style-guide#spacing-and-indentation and update the code to match our conventions for wrapping long parameter lists.

src/test/scala/com/databricks/spark/avro/AvroFromSchemaTest.scala
```diff
@@ -0,0 +1,8 @@
+package com.databricks.spark.avro
```
JoshRosen: Every file must contain the Databricks license header.

src/test/scala/com/databricks/spark/avro/AvroFromSchemaTest.scala
```diff
@@ -0,0 +1,8 @@
...
+
+/**
+  * Created by ogirardot on 24/01/16.
```
JoshRosen: Please remove this comment.

src/test/scala/com/databricks/spark/avro/AvroFromSchemaTest.scala
```diff
@@ -0,0 +1,8 @@
...
+  */
+class AvroFromSchemaTest {
```
JoshRosen: I'm assuming that this PR is WIP until this has tests.

ogirardot:
@JoshRosen yep tests are in progress, but the test files (existing ones) do not have the databricks license header. Do you want me to add it to the new class or all the other ones as well ?

JoshRosen:
It's fine if you only add the header to the new class introduced in this PR; best to keep the changes here narrowly focused.

Can you compare / contrast your approach to #96 with the one in #109?

I'll let @vlyubin or another member of the team handle the actual review for this PR, but I just wanted to take an initial pass to get some of the style + test roadblocks out of the way.

ogirardot:
This PR allows you to specifically use as input/output schema an Avro class in the classpath by using the "options" map of read/write, whereas in #106 you specify the Schema itself with a dedicated method.
But in #106 there's also a handling of nulls, that I did not quite get so I did not add them to this PR.

I took this path (class-based instead of schema-based) because from my experience this is what I'd need most of the time for our projects. It can be discussed of course.

I'll fix the test.

ogirardot:
@JoshRosen done, with tests.

bottleimp:
Hi @ogirardot, @JoshRosen, I'm ok with both the class-based and schema-based way. It's just some of my projects were using `GenericRecord` rather than `SpecificRecord`, so I prefer more on the schema-based.

But there is a schema **evolution** problem which #96 mentioned I think your PR is not solved. For example, current avsc is like this:

```
{"namespace": "example.avro",
 "type": "record",
 "name": "User",
 "fields": [
     {"name": "name", "type": "string"},
     {"name": "favorite_number",  "type": ["int", "null"]},
     {"name": "favorite_color", "type": ["string", "null"]}
 ]
}
```

then for some reason, you add a new `favorite_movie` field to the avsc, so it becomes

```
{"namespace": "example.avro",
 "type": "record",
 "name": "User",
 "fields": [
     {"name": "name", "type": "string"},
     {"name": "favorite_number",  "type": ["int", "null"]},
     {"name": "favorite_color", "type": ["string", "null"]},
     {"name": "favorite_movie", "type": ["string", "null"], "default": "Star War"}
 ]
}
```

then when you try use this schema to read former avro files, you may get exception like this:

	java.lang.AssertionError: Invalid column favorite_movie

This is what my handling of null at #109 doing, although the solution is far from perfect. Mainly because I can't find a way for `org.apache.spark.sql.Row` to store avro's default value info, so currently my solution just simply make it `null`, hope someone has a better idea.

jaley:
Would it be possible to enhance this PR by also allowing the schema to be provided explicitly as a string, rather than through the generated Avro code? 

One of the primary attractions for using Avro for many users will be that static code generation is optional. At SwiftKey, we use the statically generated classes in our apps where they want the POJOs etc for runtime app logic, but our data systems don't use it. This is because we don't want to redeploy our systems with every schema update - we instead store our schema revisions in a database and retrieve the latest to create a generic record reader capable of reading all records for all versions.

So, this change would be great for us and really simplify our ETL process if there was also the ability to specify the AVSC JSON explicitly.

bottleimp:
hi @jaley, you can check this pr #109, to see whether it meets your need

jaley:
@bottleimp Ah right, yeah I missed that. Although we're using this primarily from a Python app, and IIUC your change only provides a Scala API? I think the approach of using a configuration option would be preferable from our point of view, as it's accessible from all Spark supported languages.

Seems this change and the one in #109 have some considerable overlap? Is there a way we can combine them sensibly?

ogirardot:
@jaley Hi, I added two parameters `avro.input.schema.avsc` and `avro.output.schema.avsc` and defined as default behaviour that the "class"-based parameter has precedence + rebased the work to be merge-able.

I'll see what I can do regarding the nulls.

bottleimp:
@jaley yes, you are right, #109 only support Scala now, at this point, I think @ogirardot's approach is better.

jaley:
We did a little more investigation and I think there's actually a bug relating to this change.  If I understand correctly, setting the read schema here only sets the schema being used to translate the `GenericRecord` instances to `Row`s. It doesn't set the read schema on the underlying Avro `DataFileReader`, which means that Avro is using the schema embedded in the data files to deserialize the objects.

This is problematic, as we have assumptions in the `SchemaConversions` functions that are more restrictive than Avro compatibility constraints. Specifically, we rely on the ordering of fields within records [here](https://github.com/databricks/spark-avro/blob/master/src/main/scala/com/databricks/spark/avro/SchemaConverters.scala#L152) being the same as the reader schema. Avro supports addition of new fields at any place in the schema, so long as it has a default value. This code will therefore only work when we know that the schema passed to `createConverterToSQL()` is the *same* schema as in the `GenericRecord` instances passed to the returned converter function later.

For example, try this:
* Take an Avro schema with two fields and use that to serialise some data into Avro's data file format. 
* Modify the schema to include one new field, position it in the middle of the existing two fields and give it a default value. This new schema is technically compatible with the old one, as can be confirmed with [SchemaCompatibility.checkReaderWriterCompatibility()](https://avro.apache.org/docs/1.7.7/api/java/org/apache/avro/SchemaCompatibility.html#checkReaderWriterCompatibility(org.apache.avro.Schema, org.apache.avro.Schema)).
* Use spark-avro to read that original data file, setting the input schema as the new version using the change in this PR. One of two things will happen:
  * If your new field is the same type as the originals, you'll find it contains the data of your original second field, and the new field contains null values
  * If your new field is a different type, you'll get match errors when you query the data later on as Catalyst has the wrong type information for the second column.

On the plus side, I think the solution is actually quite simple, but we should be aware of the consequences. IMHO, we shouldn't be trying to do Avro's job for it w.r.t. schema migration. We need to find a way of setting the reader schema on the Avro `DataFileReader`, which (I think) when using the map/reduce API is through [AvroJob.setInputSchema()](http://avro.apache.org/docs/1.7.7/api/java/org/apache/avro/mapred/AvroJob.html#setInputSchema(org.apache.hadoop.mapred.JobConf, org.apache.avro.Schema)). We don't actually have the Hadoop JobConf instance to hand in `buildScan()`, so that might cause problems. We tested this by manually setting our JSON schema in the Hadoop Configuration instance, using the `AvroJob.INPUT_SCHEMA` key, as that's essentially what that function does anyway, and it seems to work, but maybe somebody can suggest a better way?

Having done that, we can be sure that all `GenericRecord` instances emitted in the iterator used in `buildScan()` are already compatible and follow the same field layout as the given reader schema, so the existing conversion code all works fine. It will however mean that we inherit the default values from the schema for those new fields, which previously didn't happen. I think most would see that as a positive thing, but we should be aware that it's a change in behaviour. Additionally, I don't know if there are perhaps situations where our current implementation is *more* forgiving that Avro's compatibility constraints - that would potentially be more of a problem?

jaley:
Actually I just realised that #95 actually already contains the change to set the underlying Avro reader schema.

pauldwolfe:
@jaley How did you manage to get AvroRelation to pick up AvroJob.INPUT_SCHEMA when constructing a DataFileReader/DaumReader?  From what I can see those config options are only used in AvroDataReader which AvroRelation doesn't leverage?

Bit hard to determine what the status is here? Does #95 fix the issue? What's the plan for merging there?
.

jaley:
@pauldwolfe AvroRelation is using the Avro MapReduce classes - the subclasses of Hadoop InputFormats, along with the SparkContext.newHadoopFile() API. There's a static function we're supposed to use on AvroJob to set the read schema when using this API, but we can't do it inside buildScan() as it requires access to the Hadoop Job instance. However, looking at the implementation of [AvroJob.setInputSchema()](http://avro.apache.org/docs/1.7.7/api/java/org/apache/avro/mapred/AvroJob.html#setInputSchema(org.apache.hadoop.mapred.JobConf, org.apache.avro.Schema)), it seems all that does is sets a config option. #95 actually includes this change, [here](https://github.com/databricks/spark-avro/pull/95/files#diff-6f47945026ae49815f8c2de96529f225R136) by setting the config itself. That appears to work for me -- passing the read schema in like this does cause the right thing to happen with the GenericRecord instances passed to the row converter.

The more serious issue, IMO, addressed in #95 is the use of globs for file paths. Reading a directory with thousands of Avro files is impossible without that change, hence we're still running a fork of this repo.

pauldwolfe:
@jaley Agreed, we're planning a similar usage pattern. Would prefer not to fork if this will be merged soon (ie weeks not months). Any clue on the ETA?  Looks like your last comment around a month ago received no reply?

JoshRosen:
I think that this PR has been largely subsumed by other more recent ones and it's also pretty merge conflicted, so I'm going to close this for now. If subsets of this PR's features are still missing, then please submit a new PR. Thanks!

ogirardot:
will do thx

PR #128, opened 2016-04-25
Ports spark-avro to Spark 2.0

liancheng:
This PR ports spark-avro to the file source API in Spark 2.0. Because the file source API hasn't been marked as public API and currently relies on some internal APIs, this PR also uses internal APIs like `InternalRow` and `GenerateUnsafeProjection`. These should be removed in future versions as these are subject to change over time.


liancheng:
We should probably create a 2.0 branch and re-target this PR to it.

src/main/scala/com/databricks/spark/avro/Logging.scala
```diff
@@ -0,0 +1,90 @@
...
+package com.databricks.spark.avro
+
+import org.slf4j.{Logger, LoggerFactory}
+
+/**
+ * Utility trait for classes that want to log data. Creates a SLF4J logger for the class and allows
+ * logging messages at different levels using methods that only evaluate parameters lazily if the
+ * log level is enabled.
+ */
+trait Logging {
```
liancheng: This trait is added because `org.apache.spark.Logging` is no longer available in Spark 2.0. Suggestions about best practices for logging are welcome.

src/test/scala/com/databricks/spark/avro/AvroSuite.scala
```diff
@@ -1,19 +1,36 @@
+/*
```
liancheng: Changes in this file are mostly styling fixes.

src/main/scala/com/databricks/spark/avro/DefaultSource.scala
```diff
@@ -16,16 +16,218 @@
...
+
+          var i = 0
+          while (i < requiredSchema.length) {
+            fieldExtractors(i)(record, rowBuffer)
+            i += 1
+          }
+
+          val internalDataRow = encoder.toRow(new GenericRow(rowBuffer))
+          val withPartitionValues = joinedRow(internalDataRow, file.partitionValues)
+          toUnsafeRow(withPartitionValues)
```
liancheng: `FileFormat.buildReader()` requires returning `UnsafeRow` here.

liancheng:
Tried to run the built-in benchmarks but hit class loading issues (Scala classes couldn't be found), still need some investigation.

liancheng:
Travis builds failed because Spark 2.0.0-SNAPSHOT artifacts are missing for now.

build.sbt
```diff
@@ -2,13 +2,13 @@ name := "spark-avro"
...
 
-scalaVersion := "2.10.5"
+scalaVersion := "2.11.7"
 
 crossScalaVersions := Seq("2.10.5", "2.11.7")
 
 spName := "databricks/spark-avro"
 
-sparkVersion := "1.4.1"
+sparkVersion := "2.0.0-SNAPSHOT"
```
marmbrus: You can point to the snapshot repo.  Look at spark-sql-perf for an example.

src/main/scala/com/databricks/spark/avro/DefaultSource.scala
```diff
@@ -16,16 +16,218 @@
...
+
+      new Iterator[InternalRow] {
+        val rowBuffer = Array.fill[Any](requiredSchema.length)(null)
+
+        // Used to convert `Row`s containing data columns into `InternalRow`s.
+        val encoder = RowEncoder(requiredSchema)
+
+        val joinedRow = new JoinedRow()
+
+        val toUnsafeRow = {
```
marmbrus: can't you just use the encoder here?

marmbrus:
Can you also bump the version of the library and update the readme?

src/main/scala/com/databricks/spark/avro/DefaultSource.scala
liancheng: Ah, forgot that `RowEncoder` already outputs `UnsafeRow`s. Thanks!

liancheng:
@marmbrus Should we bump spark-avro to 3.0.0-SNAPSHOT since this branch is no longer compatible with Spark 1.x?

liancheng:
Bumped spark-avro version to 3.0.0-SNAPSHOT and updated Travis CI settings.

liancheng:
It seems that I made some mistake related to Scala 2.10. 

```
$ ./build/sbt -Dhadoop.testVersion=2.2.0 -Dspark.testVersion=2.0.0-SNAPSHOT
> ++ 2.10.5
> compile
```

Then I get exactly the same compilation error in [this failed Travis build][1]. I tried to inspect the dependencies using

```
> show dependencyClasspath
```

and found `scala-library-2.11` and `scala-reflect-2.11`. But haven't figured out why 2.11 dependencies are pulled in. Tried to inspect the dependency tree by adding the `sbt-dependency-graph` plugin, but it seems that dependencies pulled in by `sbt-spark-package` are not included:

```
> dependencyTree
[warn] Credentials file /Users/lian/.bintray/.credentials does not exist
[info] Updating {file:/Users/lian/local/src/spark-avro/}spark-avro...
[info] Resolving org.fusesource.jansi#jansi;1.4 ...
[info] Done updating.
[info] com.databricks:spark-avro_2.10:3.0.0-SNAPSHOT [S]
[info]   +-org.apache.avro:avro:1.7.6
[info]     +-com.thoughtworks.paranamer:paranamer:2.3
[info]     +-org.apache.commons:commons-compress:1.4.1
[info]     | +-org.tukaani:xz:1.0
[info]     |
[info]     +-org.codehaus.jackson:jackson-core-asl:1.9.13
[info]     +-org.codehaus.jackson:jackson-mapper-asl:1.9.13
[info]     | +-org.codehaus.jackson:jackson-core-asl:1.9.13
[info]     |
[info]     +-org.slf4j:slf4j-api:1.6.4
[info]     +-org.xerial.snappy:snappy-java:1.0.5
```

@brkyvz Did I make some wrong configuration related to `sbt-spark-package`?

[1]: https://travis-ci.org/databricks/spark-avro/jobs/125790387

build.sbt
liancheng: Thanks, it works now.

brkyvz:
@liancheng It's weird because you specified Scala version as 2.11.7, but you are pulling in Scala 2.11.8! while trying to compile for Scala 2.10. I don't see any issues with the build configuration. It's like one of the dependencies is pulling transitives.

brkyvz:
@marmbrus @liancheng @JoshRosen  I guess Spark 2.0.0-SNAPSHOTS for Scala 2.10 are being published with the wrong Scala configuration. This is inside the Ivy xml for `spark-sketch_2.10`:
```
<m:properties__scala.version>2.11.8</m:properties__scala.version>
<m:properties__scala.binary.version>2.10</m:properties__scala.binary.version>
```

Causing you to pull Scala 2.11, even for a 2.10 dependency.

JoshRosen:
Yeah, I've been meaning to fix this. The problem is that we are no longer
publishing dependency reduced POMs for Spark following the removal of the
'unused' JAR that we used to trigger the shade plugin. I'll fix this now
and will kick off a new snapshot build.

vanzin:
>  The problem is that we are no longer publishing dependency reduced POMs for Spark following the removal of the 'unused' 

Hmm, we still have the shade plugin for jetty and guava, I was under the impression that was enough? Or maybe spark-sketch does not depend on any of those and that does not trigger the shade plugin? (In that case, I think I'd prefer a dummy dependency on guava rather than re-introducing the unused jar.)

JoshRosen:
> Or maybe spark-sketch does not depend on any of those and that does not trigger the shade plugin?

Yeah, that's the problem. Since `spark-sketch` is designed for standalone consumption outside of Spark, I'm not a fan of introducing Guava bits into its build, especially since that might complicate what we have to reason about if we ever bump Guava versions. Instead, I'd rather just add back the `unused` dependency because it's easy to reason about, has a single clear purpose, and isn't that messy / confusing.

JoshRosen:
I've opened https://github.com/apache/spark/pull/12706 to fix this in Spark.

JoshRosen:
I just published new Maven snapshots and retested this; it's now failing due to a different, legitimate-looking issue.

liancheng:
@brkyvz @JoshRosen Thanks a lot for the help! The current failure is due to recent major refactoring on Spark master branch.

koertkuipers:
we build and published this inhouse without any issues, but then unit tests in our inhouse libraries that simply roundtrip avro data (write to file and read back out from file) started failing when we used this as a dependency. errors:
```
[info]   java.lang.OutOfMemoryError: Java heap space
[info]   at java.util.Arrays.copyOf(Arrays.java:2367)
[info]   at java.lang.StringCoding.safeTrim(StringCoding.java:89)
[info]   at java.lang.StringCoding.decode(StringCoding.java:230)
[info]   at java.lang.String.<init>(String.java:453)
[info]   at java.lang.String.<init>(String.java:505)
[info]   at org.apache.spark.unsafe.types.UTF8String.toString(UTF8String.java:821)
[info]   at org.apache.spark.sql.catalyst.expressions.GeneratedClass$SpecificSafeProjection.apply(Unknown Source)
[info]   at org.apache.spark.sql.catalyst.encoders.ExpressionEncoder.fromRow(ExpressionEncoder.scala:241)
[info]   at org.apache.spark.sql.Dataset$$anonfun$org$apache$spark$sql$Dataset$$execute$1$1$$anonfun$apply$13.apply(Dataset.scala:2103)
[info]   at org.apache.spark.sql.Dataset$$anonfun$org$apache$spark$sql$Dataset$$execute$1$1$$anonfun$apply$13.apply(Dataset.scala:2103)
[info]   at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:245)
[info]   at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:245)
[info]   at scala.collection.IndexedSeqOptimized$class.foreach(IndexedSeqOptimized.scala:33)
[info]   at scala.collection.mutable.ArrayOps$ofRef.foreach(ArrayOps.scala:186)
[info]   at scala.collection.TraversableLike$class.map(TraversableLike.scala:245)
[info]   at scala.collection.mutable.ArrayOps$ofRef.map(ArrayOps.scala:186)
[info]   at org.apache.spark.sql.Dataset$$anonfun$org$apache$spark$sql$Dataset$$execute$1$1.apply(Dataset.scala:2103)
[info]   at org.apache.spark.sql.execution.SQLExecution$.withNewExecutionId(SQLExecution.scala:54)
[info]   at org.apache.spark.sql.Dataset.withNewExecutionId(Dataset.scala:2390)
[info]   at org.apache.spark.sql.Dataset.org$apache$spark$sql$Dataset$$execute$1(Dataset.scala:2102)
[info]   at org.apache.spark.sql.Dataset$$anonfun$org$apache$spark$sql$Dataset$$collect$1.apply(Dataset.scala:2107)
[info]   at org.apache.spark.sql.Dataset$$anonfun$org$apache$spark$sql$Dataset$$collect$1.apply(Dataset.scala:2107)
[info]   at org.apache.spark.sql.Dataset.withCallback(Dataset.scala:2403)
[info]   at org.apache.spark.sql.Dataset.org$apache$spark$sql$Dataset$$collect(Dataset.scala:2107)
[info]   at org.apache.spark.sql.Dataset.collect(Dataset.scala:2083)
```
 
at other times it causes java to segfault.

java -version
java version "1.7.0_75"
Java(TM) SE Runtime Environment (build 1.7.0_75-b13)
Java HotSpot(TM) 64-Bit Server VM (build 24.75-b04, mixed mode)

koertkuipers:
this is the segfault i get:
```
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x00007f84661d4f54, pid=9642, tid=140206192715520
#
# JRE version: Java(TM) SE Runtime Environment (7.0_75-b13) (build 1.7.0_75-b13)
# Java VM: Java HotSpot(TM) 64-Bit Server VM (24.75-b04 mixed mode linux-amd64 compressed oops)
# Problematic frame:
# V  [libjvm.so+0x747f54]  _Copy_arrayof_conjoint_jlongs+0x44
```

koertkuipers:
disregard my previous posts (i deleted them), the issue is not specific to spark-avro but something else in spark 2.0. thanks

src/main/scala/com/databricks/spark/avro/Logging.scala
JoshRosen: I'd recommend using `slf4j-api` instead, similar to what we do in `spark-redshift`: https://github.com/databricks/spark-redshift/commit/d5fcb2975821cdf0b862faa0865c96468e845524

liancheng:
Marked as WIP. Will come back to fix the test coverage issue after finishing Spark 2.0 issues at hand, which have higher priorities.

src/main/scala/com/databricks/spark/avro/Logging.scala
liancheng: Thanks!

liancheng:
Seems that Spark 2.0.0-SNAPSHOT for Scala 2.10 is a little bit stale and doesn’t include changes made in https://github.com/apache/spark/pull/13342 Updating spark-avro and the 2.10 version complains that `DataFrameWriter` doesn’t have a type parameter. 2.11 is fine though: https://travis-ci.org/databricks/spark-avro/builds/137143803

@JoshRosen IIUC, publishing new snapshots should fix this?

JoshRosen:
@liancheng, you should be able to trigger https://amplab.cs.berkeley.edu/jenkins/view/Spark%20Packaging/job/spark-branch-2.0-maven-snapshots/ in order to build a new snapshot.

liancheng:
@JoshRosen Thanks!

liancheng:
retest this please

liancheng:
Just published another snapshot release. Local test looks good.

liancheng:
@JoshRosen Thanks! Finally passed...

src/main/scala/com/databricks/spark/avro/DefaultSource.scala
```diff
@@ -16,16 +16,210 @@
...
+import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.catalyst.InternalRow
+import org.apache.spark.sql.catalyst.encoders.RowEncoder
+import org.apache.spark.sql.catalyst.expressions.GenericRow
+import org.apache.spark.sql.execution.datasources.{FileFormat, OutputWriterFactory, PartitionedFile}
+import org.apache.spark.sql.sources.{DataSourceRegister, Filter}
 import org.apache.spark.sql.types.StructType
 
-private[avro] class DefaultSource extends HadoopFsRelationProvider {
+private[avro] class DefaultSource extends FileFormat with DataSourceRegister {
```
cloud-fan: should we rename it to `AvroFileFormat`? like the spark built-in data source do

src/main/scala/com/databricks/spark/avro/DefaultSource.scala
```diff
@@ -16,16 +16,210 @@
...
+  private val log = LoggerFactory.getLogger(getClass)
+
+  override def equals(other: Any): Boolean = other match {
+    case _: DefaultSource => true
```
cloud-fan: it's different than the previous version: https://github.com/databricks/spark-avro/pull/128/files#diff-6f47945026ae49815f8c2de96529f225L190

cloud-fan:
LGTM

src/main/scala/com/databricks/spark/avro/DefaultSource.scala
liancheng: Good point. Thanks.

liancheng: Actually we can't... `DefaultSource` is the class name that we are using to lookup concrete data source implementations. The reason why we can rename those built-in data sources is that they are [special cased][1] (which can be a little bit misleading).

[1]: https://github.com/apache/spark/blob/7b9071eeaa62fd9a51d9e94cfd479224b8341517/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala#L82-L109

liancheng: Nvm, Michael just pointed out offline that my statement above is not true.

liancheng: Tried renaming, and hit test failures.

The problem is that, if users are using fully qualified *package* name (i.e. "com.databricks.spark.avro" in the case of spark-avro), then the `DataSourceRegister` class must be named as `DefaultSource`. For backwards compatibility consideration. I'd prefer not changing it.

liancheng:
@yhuai Could you please help sign this off?

As discussed offline, we may also want to create a separate branch for existing spark-avro versions targeting Spark 1.x.

liancheng:
Just a note for future reference: Josh had reconfigured CodeCov so that the passing criterion is now "code coverage > 90%".

README.md
```diff
@@ -37,17 +42,34 @@ Using Maven:
 </dependency>
 ```
 
+### For Spark 2.0
+
+Using SBT:
+
+```
+libraryDependencies += "com.databricks" %% "spark-avro" % "3.0.0"
```
yhuai: Do we need to use `spark-avro_2.10` at here?

brkyvz: The `%%` will add the `_2.10` or `_2.11` based on your scala version of choosing

src/main/scala/com/databricks/spark/avro/DefaultSource.scala
```diff
@@ -16,16 +16,210 @@
...
+  }
+
+  override def inferSchema(
+      spark: SparkSession,
+      options: Map[String, String],
+      files: Seq[FileStatus]): Option[StructType] = {
+    val conf = spark.sparkContext.hadoopConfiguration
+
+    // Schema evolution is not supported yet. Here we only pick a single random sample file to
+    // figure out the schema of the whole dataset.
```
yhuai: We are not changing the behavior at here, right?

yhuai:
LGTM

yhuai:
Merging!

PR #130, opened 2016-05-17
convert array type instead of just casting

defg:
This is to fix a problem I was having when using https://github.com/julianpeeters/avrohugger case classes with spark. The problem stems from the fact that no native iterable scala type(List, Seq, etc) extends GenericData.Array, so using case classes for avro instead of java Specific classes doesn't work.

I'm happy to wrap this up and contribute if someone can point me to what needs to be done. If there is a better way to handle being able to use scala case classes with avro, I'm open to suggestions as well.

### DIFF
--- a/src/main/scala/com/databricks/spark/avro/SchemaConverters.scala
+++ b/src/main/scala/com/databricks/spark/avro/SchemaConverters.scala
@@ -160,7 +160,9 @@ object SchemaConverters {
         (item: Any) => if (item == null) {
           null
         } else {
-          item.asInstanceOf[GenericData.Array[Any]].map(elementConverter)
+          val converted = item.asInstanceOf[java.util.Collection[Any]]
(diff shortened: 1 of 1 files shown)


JoshRosen:
This seems like a reasonable fix on paper, but is there an easy way to write a regression test for this which demonstrates the original problem?

JoshRosen:
Okay, so I dug into this a bit and it looks like this is only relevant once you've exposed the GenericRow conversion functions (as in #89). That said, I think there's an even simpler version of this change which we can incorporate: instead of wrapping the `Collection` back into a `GenericData.Array` object, I think we can just iterate over it without re-wrapping. I'm going to open my own PR to make this smaller change.

JoshRosen:
See #194 for my version of this patch that's going to be incorporated into the next release.

PR #132, opened 2016-05-18
expose createConverterToSQL for spark-bigquery

nevillelyh:


nevillelyh:
We are using this for `RDD[GenericRecord]` -> DataFrame conversion in spark-bigquery:
https://github.com/spotify/spark-bigquery/blob/master/src/main/scala/com/spotify/spark/bigquery/package.scala#L102

JoshRosen:
Hey, so I think that adding an `RDD[AvroRecord]` to DataFrame conversion would make a bunch of sense, but I'm not sure if this is the best / cleanest way to do this, especially given the lack of user-facing documentation and the awkward interfaces / names exposed here. Given this, I'm going to have to close this PR, but I'll try to address this use-case in a more principled way by exposing more conversions with a cleaner API.

Issue #136, opened 2016-06-27
Hadoop configuration "avro.mapred.ignore.inputs.without.extension" is ignored after upgrading to Spark 2.0.0-SNAPSHOT


liancheng:
In Spark 1.x, we use `HadoopRDD` and `AvroInputFormat` to read Avro files. In this case, Hadoop configuration "avro.mapred.ignore.inputs.without.extension" is checked in `AvroInputFormat.getSplits()` to ignore files whose paths don't end with ".avro". See [here](https://github.com/apache/avro/blob/da22ffcba82efa50aa384b3acdf26561085e14f6/lang/java/mapred/src/main/java/org/apache/avro/mapred/AvroInputFormat.java#L54-L55).

However, in Spark 2.0, `FileFormat` data sources shortcut Hadoop split generation, thus this configuration is no longer checked. This causes problems when ".avro" files are mixed with other files under a single directory. For example, spark-redshift may generate ".avro" files together with a "manifest.json" file when writing a Redshift table. Now this JSON file triggers a “Not an Avro data file” error.

This issue blocks https://github.com/databricks/spark-redshift/pull/221.

rxin:
Maybe we should add a path filter as part of the API in FileFormat?

liancheng:
@rxin A path filter interface is worth adding. But it only solves this specific problem. Handling this Avro Hadoop configuration is still necessary for spark-avro. Otherwise it would be a breaking behavior change for this data source.

One take away of this issue is that we need to pay attention to logics inside those original Hadoop style `getSplits()` methods when implementing corresponding `FileFormat` data sources.

marmbrus:
+1 to adding a path filter interface.  There is also special cased code in parquet for file filtering (i.e. don't ignore _metadata files).

liancheng:
Filed https://issues.apache.org/jira/browse/SPARK-16317 for a general `FileFormat` file filtering interface.

JoshRosen:
It looks like this was fixed in #137.

PR #137, opened 2016-06-27
Respect avro.mapred.ignore.inputs.without.extension

liancheng:
This PR fixes issue #136 by checking Hadoop configuration "avro.mapred.ignore.inputs.without.extension" in `DefaultSource.buildReader()`, so that files whose paths don't end with ".avro" are ignored.

### DIFF
--- a/src/main/scala/com/databricks/spark/avro/DefaultSource.scala
+++ b/src/main/scala/com/databricks/spark/avro/DefaultSource.scala
@@ -141,52 +141,60 @@ private[avro] class DefaultSource extends FileFormat with DataSourceRegister {
       spark.sparkContext.broadcast(new SerializableConfiguration(hadoopConf))
 
     (file: PartitionedFile) => {
-      val reader = {
-        val conf = broadcastedConf.value.value
-        val in = new FsInput(new Path(new URI(file.filePath)), conf)
-        DataFileReader.openReader(in, new GenericDatumReader[GenericRecord]())
-      }
+      val conf = broadcastedConf.value.value
 
-      val fieldExtractors = {
-        val avroSchema = reader.getSchema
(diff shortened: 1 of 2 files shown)


liancheng:
cc @yhuai

clockfly:
Looks good! +1

clockfly:
By https://avro.apache.org/docs/1.8.0/api/java/org/apache/avro/mapred/AvroInputFormat.html,
Hadoop also support avro files that don't have `.avro` suffix. Do we want to keep this ability?

```
An InputFormat for Avro data files. By default, when pointed at a directory, this will silently skip 
over any files in it that do not have .avro extension. 

To instead include all files, set the avro.mapred.ignore.inputs.without.extension property
 to false.
```

marmbrus:
Why are we doing the filtering in `buildReader` instead of when we are returning lists of files.  The downside to this approach is you are still putting pressure on the scheduler for tasks that will never return any results.  This is okay for 2.0, but before we open up the FileFormat API we should probably have a more general way to solve this.

liancheng:
@marmbrus As you and @rxin commented in #136, I agree that should add a file filtering interface in `FileFormat` to fix similar issues, so that we won't generate empty tasks. This PR chose the current approach mostly because currently there's no way to hook into file listing process from the perspective of spark-avro.

liancheng:
@clockfly `IgnoreFilesWithoutExtensionProperty` is just defined as "avro.mapred.ignore.inputs.without.extension", so we do support reading Avro files that don't have ".avro" suffix with this PR.

liancheng:
Added comment for the input file filtering hack. Will merge this pending Travis since this is blocking spark-redshift migration to Spark 2.0.

PR #155, opened 2016-08-10
fix #96, support user-define schema when reading avro files

### LINKED ISSUE #96: Specifying a read schema with spark-avro

clockfly:
With this PR, we can specify a user-provided custom schema when reading avro files. The custom schema can contain non-existing fields.
 
1. If the custom schema contains non-existing field and the field is nullable, then we will fill the value as null. The non-existing fields can exists as top level columns, or nested columns.
2. If the custom schema contains non-existing field and the field is NOT nullable, then we will throw an exception.
3. If the custom schema is a subset of the avro file schema, then we will only retrieve the fields defined in custom schema.

**Example:**
```
scala> val df = Seq((1,2,3)).toDF("a", "b", "c")
scala> df.write.format("com.databricks.spark.avro").save("/tmp/output")
scala> import org.apache.spark.sql.types._

// Prepare a customized schema
scala> val struct = StructType(StructField("a", IntegerType, true) :: StructField("non_exist", IntegerType, true) :: Nil)
scala> spark.read.format("com.databricks.spark.avro").schema(struct).load("/tmp/output").show()
+---+---------+
|  a|non_exist|
+---+---------+
|  1|     null|
+---+---------+

// Save the schema to a Json string, you can later save this to a file.
scala> val jsonSchema = struct.json
jsonSchema: String = {"type":"struct","fields":[{"name":"a","type":"integer","nullable":true,"metadata":{}},{"name":"non_exist","type":"integer","nullable":true,"metadata":{}}]}

// Load the Json string back
scala> spark.read.format("com.databricks.spark.avro").schema(DataType.fromJson(jsonSchema).asInstanceOf[StructType]).load("/tmp/output").show()
+---+---------+
|  a|non_exist|
+---+---------+
|  1|     null|
+---+---------+
```

Fix #96

### DIFF
--- a/src/main/scala/com/databricks/spark/avro/SchemaConverters.scala
+++ b/src/main/scala/com/databricks/spark/avro/SchemaConverters.scala
@@ -25,7 +25,8 @@ import org.apache.avro.generic.{GenericData, GenericRecord}
 import org.apache.avro.{Schema, SchemaBuilder}
 import org.apache.avro.SchemaBuilder._
 import org.apache.avro.Schema.Type._
-import org.apache.spark.sql.Row
+
+import org.apache.spark.sql.catalyst.expressions.GenericRow
 import org.apache.spark.sql.types._
 
 /**
@@ -33,6 +34,8 @@ import org.apache.spark.sql.types._
  * versa.
  */
 object SchemaConverters {
+
+  class IncompatibleSchemaException(msg: String, ex: Throwable = null) extends Exception(msg, ex)
 
   case class SchemaType(dataType: DataType, nullable: Boolean)
 
@@ -87,11 +90,11 @@ object SchemaConverters {
             SchemaType(LongType, nullable = false)
           case Seq(t1, t2) if Set(t1, t2) == Set(FLOAT, DOUBLE) =>
             SchemaType(DoubleType, nullable = false)
-          case other => throw new UnsupportedOperationException(
+          case other => throw new IncompatibleSchemaException(
             s"This mix of union types is not supported (see README): $other")
         }
 
-      case other => throw new UnsupportedOperationException(s"Unsupported type $other")
+      case other => throw new IncompatibleSchemaException(s"Unsupported type $other")
     }
   }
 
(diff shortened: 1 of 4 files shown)


jamesmatanle:
Awesome! With #161 , great feature.

yanxiaole:
Excellent work!

src/main/scala/com/databricks/spark/avro/SchemaConverters.scala
```diff
@@ -34,6 +35,8 @@ import org.apache.spark.sql.types._
  */
 object SchemaConverters {
 
+  class InCompatibleSchema(msg: String, ex: Throwable = null) extends Exception(msg, ex)
```
liancheng: Nit: `InCompatible` => `Incompatible`

src/main/scala/com/databricks/spark/avro/SchemaConverters.scala
```diff
@@ -119,88 +122,169 @@ object SchemaConverters {
...
-        if (schema.getTypes.exists(_.getType == NULL)) {
-          val remainingUnionTypes = schema.getTypes.filterNot(_.getType == NULL)
-          if (remainingUnionTypes.size == 1) {
-            createConverterToSQL(remainingUnionTypes.get(0))
-          } else {
-            createConverterToSQL(Schema.createUnion(remainingUnionTypes))
+  def createConverterToSQL(
+      sourceAvroSchema: Schema,
+      targetSqlType: DataType)
+    : (AnyRef) => AnyRef = {
```
liancheng: Nit:

- No need for wrapping. The whole method signature fits in 100 characters.
- Parentheses around `AnyRef` is unnecessary.

src/main/scala/com/databricks/spark/avro/SchemaConverters.scala
```diff
@@ -119,88 +122,169 @@ object SchemaConverters {
...
+
+    def createConverter(
+        avroSchema: Schema,
+        sqlType: DataType, path: List[String])
+      : (AnyRef => AnyRef) = {
```
liancheng: Again, no need for wrapping, and no parentheses around `AnyRef => AnyRef`.

src/main/scala/com/databricks/spark/avro/SchemaConverters.scala
```diff
@@ -119,88 +122,169 @@ object SchemaConverters {
...
+              byteBuffer.get(bytes)
+              bytes
+            }
+
+        case (struct: StructType, RECORD) =>
+          val length = struct.fields.length
+          val converters = new Array[AnyRef => AnyRef](length)
+          val avroFieldIndexes = new Array[Int](length)
+          var i = 0
+          while(i < length) {
```
liancheng: Nit: Space after `while`.

src/main/scala/com/databricks/spark/avro/SchemaConverters.scala
```diff
@@ -119,88 +122,169 @@ object SchemaConverters {
...
+                path :+ sqlField.name)
+              converters(i) = converter
+              avroFieldIndexes(i) = avroField.pos()
+            } else if (!sqlField.nullable) {
+              throw new InCompatibleSchema(
+                s"Cannot find non-nullable field ${sqlField.name} at path ${path.mkString(".")} " +
+                  "in Avro schema\n" +
+                  s"Source Avro schema: $sourceAvroSchema.\n" +
+                  s"Target Catalyst type: $targetSqlType")
+            }
```
liancheng: Would be nice to add an explicit `else` branch for missing nullable fields.

liancheng: Would be nice to add an explicit `else` branch for missing nullable fields.

src/main/scala/com/databricks/spark/avro/SchemaConverters.scala
```diff
@@ -119,88 +122,169 @@ object SchemaConverters {
...
+            } else {
+              val byteBuffer = item.asInstanceOf[ByteBuffer]
+              val bytes = new Array[Byte](byteBuffer.remaining)
+              byteBuffer.get(bytes)
+              bytes
+            }
+
+        case (struct: StructType, RECORD) =>
+          val length = struct.fields.length
+          val converters = new Array[AnyRef => AnyRef](length)
```
liancheng: Would be better to use `Array.fill[AnyRef => AnyRef](length)(null)` for explicit array element initialization.

src/main/scala/com/databricks/spark/avro/SchemaConverters.scala
```diff
@@ -119,88 +122,169 @@ object SchemaConverters {
...
+          val avroFieldIndexes = new Array[Int](length)
```
liancheng: Would be better to use `Array.fill[Int](length)(0)` for explicit array element initialization.

src/test/scala/com/databricks/spark/avro/AvroSuite.scala
```diff
@@ -486,4 +487,44 @@ class AvroSuite extends FunSuite with BeforeAndAfterAll {
...
+        Seq(
+          StructField("non_exist_string", StringType, true),
+          StructField(
+            "record",
+            StructType(Seq(
+              StructField("non_exist_field", StringType, false),
+              StructField("non_exist_field2", StringType, false))),
+            false)))
+    val withEmptyColumn = sqlContext.read.schema(schema).avro(testFile).collect()
+    Seq(null, Tuple1(null))
```
liancheng: Missing assertion?

src/test/scala/com/databricks/spark/avro/AvroSuite.scala
```diff
@@ -486,4 +487,44 @@ class AvroSuite extends FunSuite with BeforeAndAfterAll {
...
+
+    assert(withEmptyColumn.forall(_ == Row(null: String, Row(null: String, null: String))))
+  }
```
liancheng: Let's add an assertion to check for exception thrown for missing non-nullable fields.

src/main/scala/com/databricks/spark/avro/SchemaConverters.scala
```diff
@@ -119,88 +122,169 @@ object SchemaConverters {
...
+          while(i < length) {
+            val sqlField = struct.fields(i)
+            val avroField = avroSchema.getField(sqlField.name)
+            if (avroField != null) {
+              val converter = createConverter(avroField.schema(), sqlField.dataType,
+                path :+ sqlField.name)
+              converters(i) = converter
+              avroFieldIndexes(i) = avroField.pos()
+            } else if (!sqlField.nullable) {
+              throw new InCompatibleSchema(
```
liancheng: This branch is not covered by unit test.

src/main/scala/com/databricks/spark/avro/SchemaConverters.scala
clockfly: Java will do initialization for array, not needed?

src/test/scala/com/databricks/spark/avro/AvroSuite.scala
clockfly: The Spark will explicitly call .asNullable when creating HadoopFsRelation, so this code branch cannot be covered.

See https://github.com/databricks/spark/blob/branch-2.0/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala#L419

liancheng:
Thanks, merging to master.

PR #163, opened 2016-08-14
First Record in Partition is lost in AvroUtil.convertRowRDDToRecordRDD 

alunarbeach:
When RowRDD is converted to RecordRDD, the first record in every partition is lost. This is because schema is retrieved by doing records.next().schema and the rest of the records are mapped.

Here is the fix:
```scala
  private[avro] def convertRowRDDToRecordRDD(
      rdd: RDD[Row], 
      schema: AvroSchema): RDD[GenericRecord] = {
    rdd.mapPartitions { records =>
      if (records.isEmpty) Iterator.empty
      else {
        records.map { x =>
          val convertor = SchemaConverters.createConverterToAvro(x.schema, schema)
          convertor(x).asInstanceOf[GenericRecord] }
      }
    }
  }
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,8 @@ target/
 .idea/
 .idea_modules/
 *.iml
+.cache-main
+.cache-tests
+.classpath
+.project
+.settings/
--- a/README.md
+++ b/README.md
@@ -1,3 +1,16 @@
+
+Applied patches (more or less exact): 
+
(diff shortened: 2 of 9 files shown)


JoshRosen:
It looks like this pull request contains many refactorings and changes beyond the code in either the 2.0.x or 3.0.x releases of this library, so it's not reviewable or mergable in its current form.

If this is an issue that is affecting the code in a released version then please re-open a new pull request against the appropriate branch and I'll be happy to review. Thanks!

PR #173, opened 2016-09-14
Adds support for defaults defined in user specified Avro schema

granturing:
The current Avro schema option does not seem to support defaults defined in the user-supplied schema. This is causing some confusion for users who are expecting default values but instead are getting `null`.

This change forces the user-supplied Avro schema to be used for parsing and converting Avro records to Spark SQL Rows or it'll fall-back to the inferred schema if none is supplied.

This is related to #176

### DIFF
--- a/src/main/scala/com/databricks/spark/avro/DefaultSource.scala
+++ b/src/main/scala/com/databricks/spark/avro/DefaultSource.scala
@@ -156,12 +156,20 @@ private[avro] class DefaultSource extends FileFormat with DataSourceRegister {
       ) {
         Iterator.empty
       } else {
-        val reader = {
+        val (reader, schema) = {
           val in = new FsInput(new Path(new URI(file.filePath)), conf)
-          DataFileReader.openReader(in, new GenericDatumReader[GenericRecord]())
+          val avroSchema = options.get(AvroSchema).map(new Schema.Parser().parse)
+          val datumReader = avroSchema match {
+            case Some(schema) => new GenericDatumReader[GenericRecord](schema)
+            case _ => new GenericDatumReader[GenericRecord]()
+          }
+          (DataFileReader.openReader(in, datumReader), avroSchema)
         }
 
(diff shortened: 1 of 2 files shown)


jaley:
One thing to be aware of with this feature is that data read into a Spark Dataset with a new revision of an Avro schema may not be compatible (according to Spark SQL's rules) with the same data files read using a previous revision of the schema. Avro supports type promotion and other slightly unusual compatibility features that it can do because of the requirement for the reader and writer schema to be present at read time. 

One could easily read the same data twice using spark-avro, once with a previous schema revision and then again with a later version, and the two Datasets you produce will not necessarily be unifiable. I expect some users will find this counterintuitive and raise tickets, as the errors they receive will look like a bug. 

Perhaps there's a way to make it clear that, depending on their exact user case, users will need to be careful with their choice of Avro schema evolution features?

JoshRosen:
@granturing, I'd like to make a release of this library soon and this is one of the patches that I'd like to block that on since it seems pretty important for the user-provided schema feature to be shippable. Any chance that you'll have time to address the merge conflicts here? Also, could you click the checkbox to grant maintainers access to your PR so I can make changes in your branch in case you don't have time?

JoshRosen:
@jaley, I agree that we should think carefully through the implications here for schema evolution support. +1 on adding guidelines in the README.

@yhuai @liancheng, could you help me take a look at this to figure out whether it's a "must-include" in this 3.1.0 release? I'm pretty concerned about adding support for user schema without this bugfix.

granturing:
@JoshRosen I'll take a look at it today. I checked and the allow edits from maintainers is checked.

JoshRosen:
I was able to resolve the merge conflict on the command line but it doesn't look like GitHub will let me push my changes to your fork; I think that the "allow edits from maintainers" only lets me edit via the GitHub.com editor. I'm going to go ahead and open a new PR to finish this up.

JoshRosen:
I'll pick up this work in #195.

PR #182, opened 2016-10-11
Support splitting Avro files when reading

### LINKED ISSUE #179: spark-avro 3.0 not partitioning based on hdfs blocks?
@JoshRosen 
Has something changed with spark-avro. When I ran with 1.0 and 2.0.1 versions in 1.6 I would get the number of tasks equal to number of hdfs blocks.  So for instance 600MB single avro file would get 5 tasks.  Now with 3.0 it uses one tasks to read the entire 600MB file, which is slow.

This is doing basic query like:
val  df = spark.read.format("com.databricks.spark.avro").load("file.avro");
df.filter(col("sendTime").equalTo(1450251840233226L)).map(r => r.getLong(1)).count()

koertkuipers:
Resolves #179 by allowing large Avro files to be split into multiple partitions when being read.

### DIFF
--- a/src/test/scala/com/databricks/spark/avro/AvroSuite.scala
+++ b/src/test/scala/com/databricks/spark/avro/AvroSuite.scala
@@ -43,7 +43,7 @@ class AvroSuite extends FunSuite with BeforeAndAfterAll {
 
   override protected def beforeAll(): Unit = {
     super.beforeAll()
-    spark = SparkSession.builder().master("local[2]").appName("AvroSuite").getOrCreate()
+    spark = SparkSession.builder().master("local[2]").appName("AvroSuite").config("spark.sql.files.maxPartitionBytes", 1024).getOrCreate()
   }
 
   override protected def afterAll(): Unit = {
@@ -543,5 +543,17 @@ class AvroSuite extends FunSuite with BeforeAndAfterAll {
     val withEmptyColumn = spark.read.schema(schema).avro(testFile).collect()
 
     assert(withEmptyColumn.forall(_ == Row(null: String, Row(null: String, null: String))))
+  }
+
+  test("read avro file partitioned") {
+    TestUtils.withTempDir { dir =>
+      val sparkSession = spark
+      import sparkSession.implicits._
+      val df = (0 to 1024 * 3).toDS.map(i => s"record${i}").toDF("records")
+      val outputDir = s"$dir/${UUID.randomUUID}"
+      df.write.avro(outputDir)
+      val input = spark.read.avro(outputDir)
+      assert(input.collect.toSet.size === 1024 * 3 + 1)
+    }
   }
 }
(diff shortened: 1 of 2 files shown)


yuryn:
it looks good. I have do this independently yesterday. I was inspired by AVRO implementation in [AvroRecordReader](https://github.com/apache/avro/blob/33d495840c896b693b7f37b5ec786ac1acacd3b4/lang/java/mapred/src/main/java/org/apache/avro/mapred/AvroRecordReader.java)

koertkuipers:
thanks @yuryn

sunilsbjoshi:
is this fixed and pushed ? Can I know the release version please.

Thanks.

tgravescs:
if this looks good can it be merged, what is eta?  thanks.

tgravescs:
ping, this package is pretty much useless without this fix with spark 2.x

src/test/scala/com/databricks/spark/avro/AvroSuite.scala
```diff
@@ -544,4 +544,16 @@ class AvroSuite extends FunSuite with BeforeAndAfterAll {
...
+
+  test("read avro file partitioned") {
+    TestUtils.withTempDir { dir =>
+      val sparkSession = spark
+      import sparkSession.implicits._
+      val df = (0 to 1024 * 3).toDS.map(i => s"record${i}").toDF("records")
+      val outputDir = s"$dir/${UUID.randomUUID}"
+      df.write.avro(outputDir)
+      val input = spark.read.avro(outputDir)
+      assert(input.collect.toSet.size === 1024 * 3 + 1)
```
ash211: This tests that the read record count is the same as the written record count, but not that the read Dataset is split into multiple partitions.

Can you add something like this that would confirm that as well?

`assert(input.rdd.partitions.length > 1)`

ash211:
Thanks!

JoshRosen:
I'm going to review this now. I edited the PR description to clean it up since it will become the commit message. Here's the ommitted text (for posterity)

> it would be good is someone that knows avro can check my usage of DataFileReader.sync and DataFileReader.pastSync. i dont want to make an off-by-one error. 
> i ran a few hundred tests with randomly generated avro files and checked they read back in correctly partitioned, and they did, but still...

JoshRosen:
LGTM, so I'm going to merge this into master for inclusion in the 3.1.0 release.

tgravescs:
thanks for getting this in, do you know when you will build a package that contains this?

JoshRosen:
@tgravescs, I'm planning to do this once two other open PRs get merged; I'd like to block on the union type support (since that seems to be a big blocker to being able to read certain schemas and has also been used in production already) and on a patch which I think is necessary in order for the new user-provided schema feature to be shipped in a more complete form.  You can track progress towards the release via the 3.1.0 GitHub milestone: https://github.com/databricks/spark-avro/milestone/3

In case it would be helpful, I'd be glad to add instructions on using JitPack-based SNAPSHOT builds to our README, similar to what I did in `spark-redshift`: https://github.com/databricks/spark-redshift#snapshot-builds

JoshRosen:
Just released this in 3.1.0: https://github.com/databricks/spark-avro/releases/tag/v3.1.0

tgravescs:
thanks!

PR #190, opened 2016-11-09
Fix ClassCastException when deserializing Map field

tgibson11:
My Avro object has a field defined as java.util.Map.  In that case, item ends up being an instance of MapWrapper, which cannot be cast to a HashMap.  Casting to Map works fine, and presumably would still work in any case that you could cast to a HashMap.  I see nothing here that would specifically require the HashMap implementation.

### DIFF
--- a/src/main/scala/com/databricks/spark/avro/SchemaConverters.scala
+++ b/src/main/scala/com/databricks/spark/avro/SchemaConverters.scala
@@ -16,7 +16,6 @@
 package com.databricks.spark.avro
 
 import java.nio.ByteBuffer
-import java.util.HashMap
 
 import scala.collection.JavaConversions._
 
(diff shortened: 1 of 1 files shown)


JoshRosen:
This seems fine to me, but would it be possible to add a regression test so this doesn't break in the future?

tgibson11:
I don't really have time to create a test right now, but here's how it could be done:

The current tests pass because they always populate the Avro object by reading from a file, and I'm sure a HashMap is used there (as it should be).

Once you've got the object, replace the "simple_map" field with some other type of map.  I was using a Scala map and JavaConverters, which results in a MapWrapper.  But I'm sure a TreeMap or anything else _that doesn't extend HashMap_ would work too.

Then try to write, and the current code will throw a ClassCastException.

JoshRosen:
Fair enough. I'm going to merge this now and I've added a `regression-test-needed` label to remind me to follow up later to add a test if I have time.

PR #192, opened 2016-11-22
Add tests for UNION support documented in README

JoshRosen:
This patch adds three new test cases for the `UNION` types explicitly mentioned in the README.


JoshRosen:
Going to merge this in now so that we have coverage for this. Later we can revisit refactoring these tests to remove some of the existing (and increased) duplication.

Issue #176, opened 2016-09-17
Default values from user-specified Avro schema are ignored


granturing:
The current `avroSchema` option is ignoring default values for columns. This is useful when needing to handle schema evolution with older data.

For example, currently, any missing columns simply return `null` vs the specified default:

```
val initial = Seq(("test")).toDF("existing_field")

initial.write.format("com.databricks.spark.avro")
  .mode(SaveMode.Overwrite)
  .save("/temp/default_schema_test.avro")

initial.show()
```

```
+--------------+
|existing_field|
+--------------+
|          test|
+--------------+
```

```
val avroSchema = """
{
 "type": "record",
 "name": "my_schema",
 "fields": [
   {
     "name": "existing_field",
     "type": "string"
   },
   {
     "name": "missing_field",
     "type": "string",
     "default": "hello"
   }
 ]
}
"""

val newVersion = sqlContext.read.format("com.databricks.spark.avro")
  .option("avroSchema", avroSchema)
  .load("/temp/default_schema_test.avro")

newVersion.show()
```

```
+--------------+-------------+
|existing_field|missing_field|
+--------------+-------------+
|          test|         null|
+--------------+-------------+

```

Whereas, the expected output should be:

```
+--------------+-------------+
|existing_field|missing_field|
+--------------+-------------+
|          test|        hello|
+--------------+-------------+
```

koertkuipers:
i believe whats going on is worse than just missing defaults. it seems the spark-avro codebase does not rely on avro for schema resolution at all. this means all features of avro which depend on translating data from writer's schema to reader's schema will not work.

itaysk:
@koertkuipers Thanks this. I've been struggling with an issue for too long before looking for issues and arriving here. Is this why the following doesn't work?
http://stackoverflow.com/questions/40405473/skipping-fields-in-a-record-using-spark-avro
do you know if it's planned to support schema evolution, resolution?

JoshRosen:
This is fixed by #195.

PR #195, opened 2016-11-27
Add support for defaults defined in user-specified Avro schema

JoshRosen:
This PR is based on @granturing's #173, with merge-conflicts fixed by @JoshRosen; the original description is as follows:

> The current Avro schema option does not seem to support defaults defined in the user-supplied schema. This is causing some confusion for users who are expecting default values but instead are getting `null`.
>
> This change forces the user-supplied Avro schema to be used for parsing and converting Avro records to Spark SQL Rows or it'll fall-back to the inferred schema if none is supplied.
>
> This is related to #176


JoshRosen:
This LGTM, so I'm going to merge it in and will cut a release.

PR #197, opened 2016-11-27
Use user-provided avroSchema during writes

### LINKED ISSUE #52: Save DF with specific Avro schema

JoshRosen:
This patch fixes #52 by honoring the `avroSchema` setting during writes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,7 +159,7 @@ val df = spark.read
 df.filter("doctor > 5").write.format("com.databricks.spark.avro").save("/tmp/output")
 ```
 
-You can specify a custom Avro schema:
+You can specify a custom Avro schema for use during reads or writes:
 
 ```scala
 import org.apache.avro.Schema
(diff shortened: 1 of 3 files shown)


JoshRosen:
I imagine that one form of schema evolution that a user might want to do is to work with data which conforms to a newer version of a schema but output data using the old schema.

What actually ends up happening here is that our write path constructs a `Record` using the Spark-generated schema, then passes that record to the Avro writer using the user-specified schema.

My understanding is that Avro's schema evolution rules governing conversion between these schemas will come into play at this boundary. If we were to take the user-provided schema and use it at the time that we construct the `Record` then I think we'd quickly run into trouble when the Spark `StructType` doesn't conform to the user-provided schema. But here we're guaranteeing that the `StructType` and `Record` schemas match up and `Record.schema` <=> `User Schema` mismatches should go through Avro schema evolution.

Maybe this needs more design, though: in the case where you're reading newer data using an old schema there's a pretty clear path for how to handle this, resolve fields, etc. but it's not clear to me how the write path of this is going to work when / if the field names in the user schema and Spark schema are completely disconnected and there's lots of implicit behavior coming into play. Maybe we should hold off on this until we understand the spec better and can figure out whether we want to make strict vs. lax write modes.

PR #124, opened 2016-04-08
add support for DateType

nihed:
Hi, 
based on this issue https://github.com/databricks/spark-avro/issues/67
I create this pull request

PR #209, opened 2017-01-27
Add support for DateType

vlyubin:
https://github.com/databricks/spark-avro/pull/124 rebased with some extra tests

### DIFF
--- a/src/test/scala/com/databricks/spark/avro/AvroSuite.scala
+++ b/src/test/scala/com/databricks/spark/avro/AvroSuite.scala
@@ -19,7 +19,7 @@ package com.databricks.spark.avro
 import java.io._
 import java.nio.ByteBuffer
 import java.nio.file.Files
-import java.sql.Timestamp
+import java.sql.{Date, Timestamp}
 import java.util.UUID
 
 import scala.collection.JavaConversions._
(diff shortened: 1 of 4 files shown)


src/test/scala/com/databricks/spark/avro/AvroSuite.scala
```diff
@@ -297,6 +298,25 @@ class AvroSuite extends FunSuite with BeforeAndAfterAll {
...
+      val rdd = spark.sparkContext.parallelize(Seq(
+        Row(1f, null),
+        Row(2f, new Date(1451980800000L)),
+        Row(3f, new Date(1460098800000L))
+      ))
+      val df = spark.createDataFrame(rdd, schema)
+      df.write.avro(dir.toString)
+      assert(spark.read.avro(dir.toString).count == rdd.count)
+      assert(spark.read.avro(dir.toString).select("date").collect().map(_(0)).toSet ==
+        Array(null, 1451980800000L, 1460098800000L).toSet)
```
vlyubin: I'm not familiar with java.sql.Date, but it seemed to be tossing away more than just miliseconds - e.g. when I added seconds part to these timestamps, the test no longer worked. Not sure why ...

nihed: It's a problem relative to timezone 
[info] - Date field type *** FAILED ***
[info]   Set(1451948400000, 1460066400000, null) did not equal Set(null, 1451980800000, 1460098800000) (AvroSuite.scala:315)

I will investigate

nihed: Hi, 
The issue is relative to the timestamp, We are managing Date so there is a round done to keep only the day at 00:00:00 in your local time zone 

How I fixed : 
      TimeZone.setDefault(TimeZone.getTimeZone("UTC"))
      val rdd = spark.sparkContext.parallelize(Seq(
        Row(1f, null),
        Row(2f, new Date(1451948400000L)),
        Row(3f, new Date(1460066400500L))
      ))
      val df = spark.createDataFrame(rdd, schema)
      df.write.avro(dir.toString)
      assert(spark.read.avro(dir.toString).count == rdd.count)
      assert(spark.read.avro(dir.toString).select("date").collect().map(_(0)).toSet ==
        Array(null, 1451865600000L, 1459987200000L).toSet)

vlyubin:
https://github.com/databricks/spark-avro/pull/124 has the changes

PR #206, opened 2016-12-31
Upgrade to spark 2.1.0

koertkuipers:
Closes #205 

I am not entirely sure why i had to make changes in getDefaultWorkFile. Without these changes one extra (and unnecessary) layer of subdirectories was created upon writing, which meant no avro files were found upon reading.

Issue #169, opened 2016-09-06
Add support for DataSets of Avro records


Zariel:
It would be very useful to be able to load an Avro DataSet and have the specific records, eg

```
val ds = sqlContext.read.load("path").as[SomeAvroType]
```

bdrillard:
I've prepared a solution for this request. I've added an object, [AvroEncoder](https://github.com/bdrillard/spark-avro/blob/master/src/main/scala/com/databricks/spark/avro/AvroEncoder.scala), which produces an SQL Encoder for Avro objects of a given Schema or class, and supports creation of DataSets for both SpecificRecord and GenericData.Record. We can create DataSets of objects that implement IndexedRecord by calling one of the following methods:

```
val encoder = AvroEncoder.avro[SomeAvroType](classOf[SomeAvroType]) // for SpecificRecord
val encoder = AvroEncoder.avro[SomeAvroType](someAvroSchema) // an alternative using the class schema
val encoder = AvroEncoder.avro[GenericData.Record](someAvroSchema) // for GenericData.Record
```

Then we can provide the encoder as the argument of ``as``:

```
val ds = sqlContext.read.load("path").as(encoder)
```

There's a caveat though: my implementation depends on the [ObjectType](https://github.com/apache/spark/blob/v2.0.0/sql/catalyst/src/main/scala/org/apache/spark/sql/types/ObjectType.scala#L22) class in Catalyst, which is currently private. In the mailing list the Spark folks said they'll accept a [pull-request](https://github.com/apache/spark/pull/15453) to make the class public. In the mean time, I add a packaged jar of Catalyst from my own [fork](https://github.com/bdrillard/spark/tree/branch-2.0) of the Spark branch-2.0 repo to the classpath of my spark-avro build, and the AvroEncoder works as expected.

Also there's no current implementation for the Avro Map type yet. Catalyst has a function [ExternalMapToCatalyst](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala#L547) that could probably be used to provide that functionality, but it's not available in Spark 2.0.x yet, so I wasn't able to build with it.

All the other Avro types serialize/deserialize though! Hope you find this useful in the interim.

JoshRosen:
@bdrillard, given that `ObjectType` now appears to be public in Spark, would you be interested in submitting a pull request to contribute your custom encoder to this project?

JoshRosen:
[...]

PR #212, opened 2017-02-07
Add support for Spark 2.1.x (while retaining Spark 2.0.x support)

JoshRosen:
This patch builds on #206 in order to restore support for Spark 2.0.x. This means that a single binary artifact can be used with both Spark 2.0.x and 2.1.x, simplifying the builds of downstream projects which are compatible with both Spark versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,20 @@ before_cache:
   - find $HOME/.sbt -name "*.lock" -delete
 matrix:
   include:
+    # ---- Spark 2.0.x ----------------------------------------------------------------------------
+    # Spark 2.0.0, Scala 2.11, and Avro 1.7.x
+    - jdk: openjdk7
+      scala: 2.11.7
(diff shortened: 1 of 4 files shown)


bdrillard:
A pull request for [#169](https://github.com/databricks/spark-avro/issues/169) would require the latest changes in the ObjectType API, which are only reflected in Spark 2.1.x. I think that would mean that a version of Spark-Avro with code for #169 merged, but without explicitly Spark 2.1.x, would break if the user tried to use the feature for datasets.

JoshRosen:
@bdrillard, my plan was to release a cross-compatible version as 3.2.0 now and then to revisit dropping 2.0.x compatibility and bumping the version number to 4.0.0 in master. In fact, let me go ahead and merge @koertkuipers's PR into master and then fork a branch-3.2 which adds my compatibility changes.

bdrillard:
@JoshRosen that makes sense to me. I'll prepare a PR for [#169](https://github.com/databricks/spark-avro/issues/169), and I suppose we can make sure it gets merged into a branch/release that ensures the correct version of Spark.

PR #215, opened 2017-02-08
AvroEncoder for Datasets of Generic/Specific Avro Records

bdrillard:
This PR adds an AvroEncoder, similar to the JavaBeanEncoder already in the SparkSQL api. Usage of this encoder is described in #169


bdrillard:
Some test cases in this pull-request make use of a few Avro SpecificRecords generated from an avsc file (called specific.avsc) using avro-tools-1.7.7. It looks like a Travis build where Avro 1.8.0 is used fails because the generated Java classes don't override an expected method from the 1.8.x api. Building the same records against avro-tools-1.8.1 similarly fails to build due to un-implemented methods that Avro-1.7.7 expects. Is there a way perhaps to profile the build to make use of separate versions of the generated Java classes during tests? I'm hesitant to simply throw out the tests that test the encoder over instances of SpecificRecords.

themodernlife:
@bdrillard those issues are addressed in #217, at least for the tests.

bdrillard:
See #217 for a workable form of this PR.

bdrillard:
Closing this PR in favor of #217

PR #217, opened 2017-02-10
Close #169 and add support for DataSet of Avro records

### LINKED ISSUE #169: Add support for DataSets of Avro records

themodernlife:
@bdrillard you did an amazing job getting this working and I hope you'll have a look at the slight refactoring I've done to get this over the finish line:

- Works with multiple versions of Spark (2.0.x and 2.1.x so far)
- Travis test matrix should be updated appropriately
- Works with multiple versions of Avro

The main challenge was getting past Spark version incompatibilities.  I've done that through the addition of SBT modules, liberal use of reflection and some straight up copy/paste.  I've tried to put proper fences around the hacks.

Let me know what you guys think!  Again, all credit to @bdrillard since this is 90% his work.

Cheers!


bdrillard:
@themodernlife thanks for putting this pull-request together, I certainly didn't have the SBT expertise necessary to handle multiple builds of Spark, so this additional work is just what the encoder needed. I had a look at the inlining of additional object types that aren't common to both 2.0.0 and 2.1.x, as well as the serializing of schemas (when used in parsing complex unions), those changes look fine to me. I noticed just a few stray commented-out lines from what may have been some refactoring experiments and a few style points in doc comments. Otherwise this looks great, thanks for the additional legwork on this issue.

gcpagano:
I'm trying to add the following test based on this PR, but I get a `org.apache.spark.SparkException: Task not serializable` caused by `Cause: java.io.NotSerializableException: org.apache.avro.Schema$RecordSchema`
Here the full stack trace https://gist.github.com/gcpagano/e95a351dc75c4a6ecd2a50fdd7665e90

I'm not sure if this use case wasn't meant to be supported in this PR or maybe I'm doing something wrong in the test code.

```
  test("create Dataset from GenericRecord") {
    val sparkSession = spark
    import sparkSession.implicits._

    val schema: Schema =
      SchemaBuilder
        .record("GenericRecordTest")
        .namespace("com.databricks.spark.avro")
        .fields()
        .requiredString("field1")
        .endRecord()

    implicit val enc = AvroEncoder.of[GenericData.Record](schema)

    val genericRecords = (1 to 10) map { i =>
      new GenericRecordBuilder(schema)
        .set("field1", "field-" + i)
        .build()
    }

    val rdd: RDD[Record] = sparkSession.sparkContext
      .parallelize(genericRecords)

    val ds = rdd.toDS()

    assert(ds.count() == genericRecords.size)
  }```

**Comment by themodernlife** [comment]
Hmm... ran into this with `SpecificRecord` as well.  The problem is `toDS` runs the encoder in `mapPartitions`.  In `mapPartitions` we check that the encoder's entire expression tree is serializable.  Some of the nodes in the expression tree hold a reference to the Avro `Schema` which is not, and you get errors like the above.

I'm looking into a way to solve this.  I have a fix for your unit test, but looking at the code I'm worried there could be other runtime errors scattered about.

**Comment by themodernlife** [comment]
@gcpagano pushed change for your unit test and added coverage for fixed and enum cases.  Let me know if you have any feedback.

**Comment by gcpagano** [comment]
LGTM

**Comment by tkinz27** [comment]
Been watching this PR for a while. Any reason to hold back on this issue?

**Comment by bdrillard** [comment]
@themodernlife Could you check and resolve this last merge conflict? It appears to be pretty minor.
@JoshRosen Would it be possible for us to merge this PR soon?

**Comment by themodernlife** [comment]
@bdrillard ok merge conflicts should be fixed now.  LMK any other feedback.

**Comment by rbrush** [comment]
FWIW, we've been using this branch for months to bring a large collection of complex Avro records (hundreds of fields nested several layers deep) into Spark, and it has worked great. We'd love to see this in spark-avro properly.

**Comment by mnarrell** [comment]
+1.  This would be a very welcomed addition.

**Comment by bellemare** [comment]
Would love to see this one get in soon. Is there anything blocking this moving forwards? Do we need to nag someone on this?

**Comment by AnmolMago** [comment]
This would be really great to have!
@yhuai @rxin

**Comment by racc** [comment]
+1 for merge

**Comment by bellemare** [comment]
It seems to me that Databricks doesn't really care about Avro support that much, which is a shame as we move towards a full streaming-based world. I have since moved on to alternative solutions as it's clear they don't care to support this project. Last commit merged is back in February.

**Comment by mnarrell** [comment]
Unfortunately, I cannot agree with you more.  There are several feature-rich pull requests that are tested, commented on, and simply waiting to be merged that would continue the value of this project.  This has languished and, like many others, we've abandoned this library.

**Comment by bellemare** [comment]
@JoshRosen - If you are not maintaining this project anymore, can you indicate as such and ideally hand it off to someone else?

**Comment by racc** [comment]
@bellemare, @mnarrell what alternatives to spark-avro are you using?

**Comment by marianokamp** [comment]
As an aside, why is this is not part of Spark, like Parquet? This is not a loaded question. I really want to know. 

And there is no Avro alternative or is there? I mean something that works for large batch files, for web service streams, binary, with external schema, schema evolution, usable for single records processing like in streaming. Would you guys use Thrift or Protocol Buffers if it would be just about streaming then?

I am currently evaluating options for an enterprise customer and that Avro is not part of Spark also may mean that you don’t get support. Say you bought an Hadoop distribution from a vendor and then run into a production issue the vendor may well say that this part is not what they signed up for :(

I hope it’s ok to discuss this here.

**Comment by mnarrell** [comment]
@racc No, alternatives to Avro.  We don't use structured streaming, or at least dataframes with high speed data.  We'll tolerate RDD[T] there.  Data at rest is parquet.

**Comment by racc** [comment]
@mnarrell ok well I'm interested in this PR because it fixes this issue:
https://stackoverflow.com/questions/36532237/infinite-recursion-in-createdataframe-for-avro-types

Basically you can't bean encode an RDD of an Avro Generated class to a Dataset because of the getSchema() method. Much of the data we work with gets transformed to Avro Generated classes because it's convenient to write an AVDL and then generate a Hive table from the Avro schema - keeps defining the data model in one place.

That's why I was asking about alternatives to spark-avro for encoding Avro data :)

**Comment by rxin** [comment]
Out of curiosity, why do we use avro formats rather than a more modern one like Parquet? Just asking so I can better understand the use cases.

**Comment by racc** [comment]
@rxin Mainly for the convenience use-case described above. Generating a class from an Avro Schema (AVDL) means we can parse our data into the generated POJOs and then write them out to whatever format (Avro, Parquet) etc. Since a Hive table can be created from an Avro Schema, we don't have to maintain a Hive DDL when the schema changes.
If there's a more convenient alternative to using Avro generated classes I'm open to hearing about it.

I think that Avro is used in conjunction with Kafka too. People tend to send messages in the Avro format and then use a Schema registry which defines the types of messages being sent (https://github.com/confluentinc/schema-registry).

One other minor advantage of the Avro format is that you can append data to the file - which helps mitigate the small files problem on the HDFS. Otherwise with Parquet you'd probably need to do some periodic consolidation of the files.

**Comment by tkinz27** [comment]
@rxin to add our use case in, we use avro over kafka, and this PR lets us easily work with structured streaming to read from kafka and write to our hdfs cluster. We have a secondary etl process that consolidates the avro files into parquet. Being able to append data is a big reason why we use avro.

**Comment by rbrush** [comment]
@rxin In addition to the above use cases, we have several petabytes of Avro files and many dozens of jobs using them (including use of specific records). This Avro encoder allows us to migrate to Spark datasets incrementally, which is important because a single, large move isn't doable in this case.

(I work with @bdrillard, who did the bulk of the work on this PR. We're using it in an internal fork in production and it's working well.)

**Comment by marmbrus** [comment]
Sorry for the delay!  I apologize that this repo has been ignored for a few months.  It is not abandoned and we realize that avro support is important.  Thanks for voicing your concerns!

We have made sure to devote resources in Q4 to merging patches and making a new release of this library. In particular, integration with structured streaming is important to me personally.

As an aside, you should of course benchmark your own workloads, but in all of [our experiments](https://databricks.com/blog/2017/06/06/simple-super-fast-streaming-engine-apache-spark.html) we have seen Spark structured streaming performs multiple-orders of magnitude faster than the equivalent Kafka queries. So while I love using Kafka as a central message bus, I think there is still a lot of reasons to use Spark for stream and batch processing :)

**Comment by bellemare** [comment]
@marmbrus Glad to see that my email to the Databricks support staff got some attention. I am pleased to hear that your team will be devoting some resources to improving Avro support going forwards.
Though I appreciate the benchmark link, I'll just say that functionality trumps performance; There is a reason why we aren't all writing our web apps in assembly.  It is easy for us to sacrifice performance for features and usability. Now, if Databricks can put together a streaming experience that handles Avro data seamlessly and still provides this multiple-of-magnitude performance gains then you would definitely be on the right track as far as I am concerned.

**Comment by marmbrus** [comment]
@bellemare totally agree. One of my favorite CS professors used to say "the biggest performance increase is when it goes from not working to working".  Do keep letting us know where we are falling short on required features!

Regarding the benchmark, I'm hoping to post the full code that you can run in community edition (or on a real cluster if you are a paying customer) soon.  Stay tuned to our blog.  Sorry again for the delay!

**Comment by racc** [comment]
@marmbrus Do you think that some of these changes will be backported to the previous versions of spark-avro (e.g. ver 2.0.1)? The reason I say this is that we're using Cloudera's Hadoop and they're still using Spark 1.6... which means that we're still on Spark 1.6 :\

**Comment by rxin** [comment]
Sorry definitely no. You can ask your vendor to do that, or pick a different one.

**Comment by gengliangwang** [comment]
@themodernlife could you make a clean patch?

**Comment by cloud-fan** [comment]
I'm a little confused by the PR title, why we need to add Dataset support for avro data source? Ideally a data source should return its data as a `DataFrame` and users can do `df.as[XXX]` to convert the `DataFrame` to a `Dataset` of any expected types.

Also there is no such a public API to allow users to get a typed `Dataset` from a data source, but only `DataFrame`.

**Comment by cloud-fan** [comment]
It sounds like we want a custom encoder for avro specific classes so that we can convert a `DataFrame` to a `Dataset` of avro classes.

I think a cleaner way to do it is to register these avro specific classes as UDT and define the custom logic of encoding/decoding.

**Comment by marmbrus** [comment]
@cloud-fan I don't really think that that is a good solution.  The UDT API is private and that sounds like a lot of unnecessary boiler plate to have to manually write code for each class.

It's not only useful to convert to a `Dataset[AvroRecord]` but also to take Avro encoded data (maybe from something like Kafka) and convert it to a `DataFrame`.  An `Encoder` seems like a good way to do this.

**Comment by themodernlife** [comment]
Hey everyone, it’s been a while since I’ve looked at or used this code... do we still like the approach I’ve taken with the build file to support multiple spark versions? If so I can find some time to clean up the patch so it merges clean with latest updates.

**Comment by bdrillard** [comment]
Seems like there may be some confusion as to the content of this PR. The purpose of this PR is to provide exactly the `Encoder` for both `SpecificRecord` and `GenericRecord` that @cloud-fan and @marmbrus suggest.

I had initially written the `AvroEncoder` class as part of a previous PR, #215, and @themodernlife made necessary additions/refactoring in this PR to get the enhancement to work against different versions of Spark because the `OutputFactoryWriter` API had changed between 2.0.x and 2.1.x.

The [AvroEncoder](https://github.com/databricks/spark-avro/pull/217/files#diff-de6c0e759124917fdef58c8f3d2a7799R49) class builds the necessary serialization/deserialization expressions for Avro records by traversing the Avro class or schema in a similar manner as Spark currently does for [Java Bean](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/Encoders.scala#L142) classes. So to @marmbrus's point, the `AvroEncoder` in this PR handles providing an `Encoder` for _any_ Avro Specific/Generic record without the need for the user to implement any custom serialization/deserialization logic.

Reference the added [tests](https://github.com/databricks/spark-avro/pull/217/files#diff-046473ef957615b2f0b39323e56be515R752) to see how these encoders are instantiated on both a `SpecificRecord` class/schema, or a `GenericRecord` schema (**note**, for more direct testing of the encoder we use the lower-level row api in those tests). With the `AvroEncoder`, we get the higher level semantics we would expect with an`Encoder` , e.g (in some Java):

```
Encoder<MySpecificAvro> encoder = AvroEncoder.of(MySpecificAvro.class);
Dataset<MySpecificAvro> ds = spark.read()
                                  .format("com.databricks.spark.avro")
                                  .load("/path/to/my_avro")
                                  .as(encoder);
ds.map(mySpecificAvro -> {
    // operations where the MySpecificAvro type is known
}) ...
```

I hope this clarifies the content here, I'm glad to answer any other questions.

**Comment by cloud-fan** [comment]
ok then this approach LGTM. BTW do you know why this PR is against branch-3.2 instead of master?

**Comment by bdrillard** [comment]
That'll be a question for @themodernlife.

**Comment by gengliangwang** [comment]
@bdrillard Do you have time to create a clean PR based on latest master? We will keep reviewing on the changes. Thanks!

**Comment by cloud-fan** [comment]
@themodernlife I'm afraid you need to send a new PR to set the base branch to master...

**Comment by themodernlife** [comment]
@cloud-fan I was able to update the base branch to master.  Looking into why Travis is complaining.

**Comment by gengliangwang** [comment]
Hi @themodernlife you changes contains commits from @JoshRosen and @liancheng , which should not be related to this PR.

**Comment by rxin** [comment]
I took a look at the implementation and it depends on basically all the internals of Spark, with a lot of expressions that even do code generation. Basically this would break with every new release of Spark and isn't really maintainable.

We should really design and implement a public API in Spark for expressing encoders, and then implement that here.

**Comment by rbrush** [comment]
I discussed this with @bdrillard, who did the bulk of the Encoder implementation, and we would welcome a stable public API in Spark for custom encoders like this. We will transition this logic to that when it's available. Should we log a JIRA to the Spark project to track that?

In the meantime, we realize this taps into some Spark classes that are subject to change, and are willing to update our internal usage of it as they evolve. We realize this isn't ideal, but it provides quite a bit of value to our existing workloads for reasons described earlier in this chain. Hopefully this state of affairs is temporary and we'll be able to migrate to the public API when available.

**Comment by rxin** [comment]
Certainly you can use this if you are willing to pay the cost to upgrade it with each version of Spark. But if that's the case, perhaps just do it in your fork for now?

**Comment by rbrush** [comment]
Sure, we're happy to continue with our internal fork for the time being. But of course we're only one of several interested parties on this thread, so I'm sure a public API to support this would be appreciated in general.

**Comment by rxin** [comment]
Yup designing a public API for encoder/decoders make a lot of sense in Spark proper. We should do it!

**Comment by rxin** [comment]
@rbrush @bdrillard Aside from that, is there a way to update the patch to not rely on custom code generation? We might need to add couple small expressions to Spark to do that.

If that works, perhaps we can create a release based on that first, and then use that as the basis to create the encoder API.

**Comment by bdrillard** [comment]
@rxin, to check my understanding of your question, it depends on how broadly you mean "to not rely on custom code generation". 

The [Serializer](https://github.com/themodernlife/spark-avro/blob/7869c115d2dc4dd1216c53e13a6a38412e071a0f/src/main/scala/com/databricks/spark/avro/AvroEncoder.scala#L576)/[Deserializer](https://github.com/themodernlife/spark-avro/blob/7869c115d2dc4dd1216c53e13a6a38412e071a0f/src/main/scala/com/databricks/spark/avro/AvroEncoder.scala#L364) of the `ExpressionEncoder` API produces a custom `Expression`, and that `Expression` ultimately generates custom code for the `Projection` to `UnsafeRow` for a collection of typed objects conforming to a given Avro schema. I don't think the core task of this patch (getting typed Datasets from collections of arbitrary Avro objects) could be attained without an `Encoder`, its SerDe, and the "custom code" it would generate. As we discussed a bit above, Spark-Avro already has the ability to convert a collection of Avro objects to a Dataframe, but without the Encoder SerDe we lose static type information that makes authoring different functions in the Spark API so much easier and more reasonable.

More specifically though, the functions producing the SerDe rely on currently public functions in the existing Spark expressions API, e.g. `Invoke`, `InvokeStatic`, `Literal`, etc. 

We do however, have a few functions that are custom extensions of `Expression`, namely [ObjectCast](https://github.com/themodernlife/spark-avro/blob/7869c115d2dc4dd1216c53e13a6a38412e071a0f/src/main/scala/com/databricks/spark/avro/AvroEncoder.scala#L777), [InitializeAvroObject](https://github.com/themodernlife/spark-avro/blob/7869c115d2dc4dd1216c53e13a6a38412e071a0f/src/main/scala/com/databricks/spark/avro/AvroEncoder.scala#L729), and [ExternalMapToCatalyst](https://github.com/themodernlife/spark-avro/blob/7869c115d2dc4dd1216c53e13a6a38412e071a0f/src/main/scala/com/databricks/spark/avro/AvroEncoder.scala#L113).

* ExternalMapToCatalyst was in the Spark Expressions API at the time this patch was prepared, but the function was in private scope, so we copied it inline. At this time, I believe we could opt for the version in the Spark API, and so we could remove our copy. 
* ObjectCast creates Java code for performing an inline object cast, which is necessary for instances where (in Avro for example) a getter returns only `Object` and through traversal of the schema, we know what type we want to cast the object to. I think this could be safely/easily added to the Spark API if such functionality isn't already there.
* InitializeAvroObject is exactly like [InitializeJavaBean](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/objects/objects.scala#L1229), but for creating instances of Avro objects. This function is necessary for the Deserializer of the `Encoder`, so it's not clear how we could take this function out. 

Let me know your thoughts.

**Comment by marmbrus** [comment]
@bdrillard I think it is the custom extensions of `Expression` that are the scariest as they build stringly typed code that relies on undocumented and unstable details of codegen.  If we can eliminate these, I think it significantly reduces the concern.  If we restrict ourselves to using APIs of internal `Expression`s then changes will at least give compiler errors.

Regarding the specific `Expression`s:
 - `ExternalMapToCatalyst` - catalyst is implicitly private (i.e excluded from the scala doc) so there is no need to rely on visibility modifiers.  I think we can just make this public and use it here.
 - `ObjectCast` - I don't think that we have this (but @cloud-fan might know better). If we were going to add this to Spark I might just make it an optional argument to `Invoke`.
 - `InitializeAvroObject` - This one does seem a little harder.  Again maybe we could generalize Spark's `InitializeX` to take `Seq[(String, Seq[Expression])]` or something so that it could be used for both.

**Comment by cloud-fan** [comment]
Spark doesn't have `ObjectCast`, but you can easily implement one.

**Comment by bdrillard** [comment]
@marmbrus, @cloud-fan, I'd be happy to open a PR in Spark proper with those changes, so we can then refactor this PR to make use of those public functions.

**Comment by cloud-fan** [comment]
why not just implement it here? Otherwise you can only support Spark 2.3 or later

**Comment by marmbrus** [comment]
@cloud-fan the fear is that having an external library depend on the specifics of code-gen is too brittle. I think its okay if this requires a new version of Spark if that reduces the tight coupling with the specifics of execution.  Until that is released we can update this PR by depending on a nightly snapshot and then once 2.3 is out we can make a new spark-avro release.

@bdrillard, that sounds great. Feel free to message me on the Spark JIRA / PR and I can help shepherd them in. Thanks for working on this!

**Comment by kalvinnchau** [comment]
@themodernlife 
I'm getting a `Cause: java.lang.RuntimeException: Error while encoding: java.lang.RuntimeException: Couldn't find bytes on class java.nio.ByteBuffer` Exception with this schema format:

```
{
  "type": "record",
  "name": "GenericRecordTest2",
  "namespace": "com.databricks.spark.avro",
  "fields": [
    {
      "name": "fixedUnionVal",
      "type": [
        {
          "type": "fixed",
          "name": "IPv4",
          "size": 4
        },
        "int"
      ]
    }
  ]
}
```

Here's the test I added:
```scala
  test("create Dataset from Fixed Generic Record") {
    val sparkSession = spark
    import sparkSession.implicits._

    val schema: Schema =
      SchemaBuilder
        .record("GenericRecordTest2")
        .namespace("com.databricks.spark.avro")
        .fields()
        .name("fixedUnionVal").`type`()
          .unionOf()
            .fixed("IPv4").size(4).and()
            .intType()
            .endUnion()
          .noDefault()
        .endRecord()


    println(schema)
    implicit val enc = AvroEncoder.of[GenericData.Record](schema)

    val genericRecords = (1 to 10) map { i =>
      new GenericRecordBuilder(schema)
        .set("fixedUnionVal", i)
        .build()
    }

    val rdd: RDD[GenericData.Record] = sparkSession.sparkContext
      .parallelize(genericRecords)

    val ds = rdd.toDS()

    assert(ds.count() == genericRecords.size)
  }
```

Gist to the stacktrace [https://gist.github.com/kalvinnchau/a9a2646450438d7ead88e6ea6c15221c](url)

bdrillard:
@marmbrus I've logged [SPARK-22739](https://issues.apache.org/jira/browse/SPARK-22739) to track the additions to the `Expressions` API.

@kalvinnchau Thanks for the additional test case! We'll be circling back to this once the Spark ticket I link just above is resolved.

bdrillard:
An update for this issue, I've created a PR to Spark that includes the necessary expressions for an Avro-Encoder, see [#21348](https://github.com/apache/spark/pull/21348).

cc: @marmbrus

mtraynham:
Well I think I resolved the issue, I just cast it to `CharSequence` instead of Utf8, with:

```diff
@@ -604,7 +604,7 @@ private object AvroTypeInference {
 
       ExternalMapToCatalyst(
         inputObject,
-        ObjectType(classOf[org.apache.avro.util.Utf8]),
+        ObjectType(classOf[java.lang.CharSequence]),
         serializerFor(_, Schema.create(STRING)),
         valueType,
         serializerFor(_, valueSchema),
```

There was also a Hive issue which is closely related that I kinda of stole the idea for the fix from:  https://issues.apache.org/jira/browse/HIVE-5865

Issue #92, opened 2015-10-14
spark-avro 2.0.1 generates strange schema (spark-avro 1.0.0 is fine)


alexnastetsky:
I save my dataframe to avro with spark-avro 1.0.0 and it looks like this (using avro-tools tojson):

{"field1":"value1","field2":976200}
{"field1":"value2","field2":976200}
{"field1":"value3","field2":614100}

But when I use spark-avro 2.0.1, it looks like this:

{"field1":{"string":"value1"},"field2":{"long":976200}}
{"field1":{"string":"value2"},"field2":{"long":976200}}
{"field1":{"string":"value3"},"field2":{"long":614100}}

At this point I'd be happy to use spark-avro 1.0.0, except that it doesn't seem to support specifying a compression codec (I want deflate).

alexnastetsky:
Simple repro case:

test.json:
```
{"field1":"foo","field2":123}
{"field1":"bar","field2":456}
```

(spark-shell):
```
scala> val df = sqlContext.read.json("test.json")
scala> df.write.format("com.databricks.spark.avro").save("test.avro")
```

java -jar avro-tools.jar tojson test.avro/part-r-00000-[uuid].avro 
```
{"field1":{"string":"foo"},"field2":{"long":123}}
{"field1":{"string":"bar"},"field2":{"long":456}}
```

I've tried this with both spark 1.4.0 and 1.5.1.

alexnastetsky:
The issue is that the StructFields in the schema are being treated as nullable=true, even if you pass in false. I changed them to nullable=true in spark-avro 1.0.0 and was able to see the same weird behavior. 

Does that mean this bug existed in 1.0.0 as well (albeit conditionally), or is output like this actually desirable?

alexnastetsky:
The reason it thinks it's nullable=true in 2.0.1 is because the DefaultSource class was changed from CreatableRelationProvider to HadoopFsRelationProvider, and in ResolvedDataSource#apply in the HadoopFsRelationProvider case, it uses dataSchema.asNullable to create the AvroRelation.

alexnastetsky:
Looks like 1.0.0 will suffice for me, got compression working with a simple `sc.hadoopConfiguration.setBoolean("mapreduce.output.fileoutputformat.compress",true)`

tyro89:
[...]

PR #218, opened 2017-02-13
Failing test where schema changes on write

ahobson:
This is a failing test as requested in Issue #92.


gengliangwang:
I think such case should be added in the PR that resolve the issue.
Close this one.

PR #220, opened 2017-03-14
Make "createConverterToAvro" function public static

ovj:
This will help createConverterToAvro being used outside of AvroOutputWriter.


mnarrell:
Wouldn't it make sense to change the access modifiers on all these schema conversion functions?  We have lots of AVRO encoded data that is not accessible via a filesystem path (stored in Redis/RDBMS/etc) that we'd love to transform into Dataframes and vice versa.

Issue #235, opened 2017-06-15
User defined type or enums support


Ravil-Safin:
Avro scheme supports ENUM datatype. Spark SQL through the UserDefinedType somehow [supports ](https://stackoverflow.com/questions/32440461/how-to-define-schema-for-custom-type-in-spark-sql) enums too. But currently spark-avro plugin doesn't support saving enums or userdefinedtypes to avro format. It would be good if plugin would support it.

jon-morra-zefr:
This is a must for me too.

PR #239, opened 2017-06-28
Getting user defined types working

jon-morra-zefr:
As outlined in #235 it is desirable to get Avro serialization working with `UserDefinedTypes`.  This PR addresses that concern and allows serialization of `UserDefinedTypes`.  I wasn't able to write any tests, but I'm sure that can be done easily.  This works on my local projects.

### DIFF
--- a/src/main/scala/org/apache/spark/avro/AvroOutputWriter.scala
+++ b/src/main/scala/org/apache/spark/avro/AvroOutputWriter.scala
@@ -14,31 +14,29 @@
  * limitations under the License.
  */
 
-package com.databricks.spark.avro
+package org.apache.spark.avro
 
 import java.io.{IOException, OutputStream}
 import java.nio.ByteBuffer
-import java.sql.Timestamp
-import java.sql.Date
+import java.sql.{Date, Timestamp}
 import java.util.HashMap
-
-import org.apache.hadoop.fs.Path
-import scala.collection.immutable.Map
 
 import org.apache.avro.generic.GenericData.Record
 import org.apache.avro.generic.GenericRecord
-import org.apache.avro.{Schema, SchemaBuilder}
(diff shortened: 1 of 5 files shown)


src/main/scala/org/apache/spark/avro/AvroOutputWriter.scala
```diff
@@ -14,31 +14,29 @@
  * limitations under the License.
  */
 
-package com.databricks.spark.avro
+package org.apache.spark.avro
```
jon-morra-zefr: In order to reference `UserDefinedType` we have to change the package name.

src/main/scala/org/apache/spark/avro/AvroOutputWriter.scala
```diff
@@ -87,8 +85,8 @@ private[avro] class AvroOutputWriter(
         case bytes: Array[Byte] => ByteBuffer.wrap(bytes)
       }
       case ByteType | ShortType | IntegerType | LongType |
-           FloatType | DoubleType | StringType | BooleanType => identity
-      case _: DecimalType => (item: Any) => if (item == null) null else item.toString
+           FloatType | DoubleType | BooleanType => identity
+      case _: DecimalType | StringType => (item: Any) => if (item == null) null else item.toString
```
jon-morra-zefr: My use case for this is serializing `Enum[_]` types.  I'm doing this as strings on the backend.  However, in my tests it was still presenting the native enum type to the native avro reader.  By moving the `StringType` out and explicitly calling `.toString` on it I'm able to get around this issue assuming the `UserDefinedType` for my enums is `StringType`.

zeitan:
Question, some plan to merge this improve???
I'm getting a possible problem trying to deserialize a avro enum filed in spark and I think that this is the reason

PR #242, opened 2017-07-18
Spark 2.2.0 Support

squito:
This adds support for spark 2.2.0.  Primarily this is addressing the api change introduced by [SPARK-19085](https://issues.apache.org/jira/browse/SPARK-19085) https://github.com/apache/spark/commit/b3d39620c563e5f6a32a4082aa3908e1009c17d2.  This fixes the issue in the most simplistic way: it copies the old conversion from `InternalRow` to `Row`.  A better implementation would do something more efficient with `InternalRow`.

This keeps both `write()` methods, so it should be compatible with all spark 2+ versions.

It also fixes some dependency conflicts when running tests -- it seems curator has a conflicting version of guava with hadoop, but we don't actually need curator for tests.

Tested by running unit tests locally.

Fixes https://github.com/databricks/spark-avro/issues/240

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,14 @@ matrix:
     - jdk: openjdk7
       scala: 2.10.6
       env: TEST_HADOOP_VERSION="2.2.0" TEST_SPARK_VERSION="2.1.0" TEST_AVRO_VERSION="1.8.0" TEST_AVRO_MAPRED_VERSION="1.8.0"
+    # Spark 2.2.0, Scala 2.11, and Avro 1.7.x
+    - jdk: openjdk8
+      scala: 2.11.8
+      env: TEST_HADOOP_VERSION="2.2.0" TEST_SPARK_VERSION="2.2.0" TEST_AVRO_VERSION="1.7.6" TEST_AVRO_MAPRED_VERSION="1.7.7"
+    # Spark 2.2.0, Scala 2.10, and Avro 1.8.x
+    - jdk: openjdk8
+      scala: 2.10.6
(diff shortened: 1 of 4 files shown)


squito:
hmm, I guess travis configuration needs to be changed to build with jdk 8

marcintustin:
Would be wonderful if this could be merged

marcintustin:
For now I'm running this with jitpack :(

rxin:
It is currently failing codecov ...

ritesh-dineout:
Is there any timeline by which this will be merged and this will be released? We started facing this issue in production, because of `aws emr` upgrading to spark 2.2.

squito:
code coverage is now passing.

The added test is pretty lame to get code coverage.  Probably the right thing to do is to start testing against spark 2.2, but that requires reconfiguring travis to use java 1.8, and I don't have permissions for that.

vanzin:
Isn't all travis configuration in `https://github.com/databricks/spark-avro/blob/branch-3.2/.travis.yml`?

squito:
oh good point.  I actually meant codecoverage, but taking a closer look at the build process I guess that is automatic from travis.  lets see what coverage I get from the latest changes

rxin:
Alright thanks.

vara-bonthu:
Just checking if this is expected to be released as 3.3.0 some time today? Thanks

tgravescs:
Just curious when is 3.3 expected to release?

praneetsharma:
Hi, I have a naive question on InternalRow to Row conversion. Since this conversion will happen for each row, can it result in performance degradation in case of millions of records?

febinsathar:
when will 3.3 release? thank you

gengliangwang:
@squito @marcintustin @ritesh-dineout  @febinsathar spark-avro 4.0.0 is released  :)

PR #249, opened 2017-10-12
Fix bug on generation of avro files with nested records with the same name

gsolasab:
spark-avro was failing on the generation of avro files when nested records had the same name, as can be seen on the following issue [#54](https://github.com/databricks/spark-avro/issues/54). 
For solving that problem has been added a different namespace on each record. This namespace is created with the concatenation of all the parents of each record separated by dot.
Added also a new test in AvroSuite for checking the new code.
This is related with pull request [#73](https://github.com/databricks/spark-avro/pull/73) but have been added a modification on AvroOutputWriter to solve correctly the issue.

### DIFF
--- a/src/main/scala/com/databricks/spark/avro/AvroOutputWriter.scala
+++ b/src/main/scala/com/databricks/spark/avro/AvroOutputWriter.scala
@@ -107,8 +107,7 @@ private[avro] class AvroOutputWriter(
         val elementConverter = createConverterToAvro(
           elementType,
           structName,
-          SchemaConverters.getNewRecordNamespace(elementType, recordNamespace, structName)
-          )
+          SchemaConverters.getNewRecordNamespace(elementType, recordNamespace, structName))
         (item: Any) => {
           if (item == null) {
             null
@@ -128,8 +127,7 @@ private[avro] class AvroOutputWriter(
         val valueConverter = createConverterToAvro(
           valueType,
           structName,
-          SchemaConverters.getNewRecordNamespace(valueType, recordNamespace, structName)
-          )
+          SchemaConverters.getNewRecordNamespace(valueType, recordNamespace, structName))
         (item: Any) => {
           if (item == null) {
             null
@@ -149,9 +147,7 @@ private[avro] class AvroOutputWriter(
           createConverterToAvro(
             field.dataType,
             field.name,
-            SchemaConverters.getNewRecordNamespace(field.dataType, recordNamespace, field.name)
-            )
-        )
+            SchemaConverters.getNewRecordNamespace(field.dataType, recordNamespace, field.name)))
         (item: Any) => {
           if (item == null) {
             null
--- a/src/main/scala/com/databricks/spark/avro/AvroOutputWriter.scala
+++ b/src/main/scala/com/databricks/spark/avro/AvroOutputWriter.scala
@@ -104,7 +104,13 @@ private[avro] class AvroOutputWriter(
       case DateType => (item: Any) =>
         if (item == null) null else item.asInstanceOf[Date].getTime
       case ArrayType(elementType, _) =>
-        val elementConverter = createConverterToAvro(elementType, structName, recordNamespace)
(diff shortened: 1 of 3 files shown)


lyubinetz:
Ping @JoshRosen - this seems like it actually does what https://github.com/databricks/spark-avro/pull/73 had to accomplish. A couple of other people asked for a solution to this, so since this is a relatively common issue (and kind of a bug), this should probably be merged.

P.S. This is @vlyubin from different account :)

gengliangwang:
retest this please

bloomonkey:
We're getting bitten by this too. If we (@autotraderuk) were able to contribute the missing tests, how soon could the fix be released?

gsolasab:
Tomorrow I will do the tests.

gengliangwang:
@gsolasab  Cool, please add test cases on file reading as well.
We plan to release recently, and I think we should merge this PR.

src/test/scala/com/databricks/spark/avro/AvroSuite.scala
```diff
@@ -693,4 +691,65 @@ class AvroSuite extends FunSuite with BeforeAndAfterAll {
...
+  test("saving avro that has nested records with the same name") {
+    TestUtils.withTempDir { tempDir =>
+      // Save avro file on output folder path
+      val writeDf = spark.createDataFrame(List(NestedTop(1, NestedMiddle(2, NestedBottom(3, "1")))))
+      val outputFolder = s"$tempDir/duplicate_names/"
+      writeDf.write.avro(outputFolder)
+      // Read avro file saved on the last step
+      val readDf = spark.read.avro(outputFolder)
+      // Check if the written DataFrame is equals than read DataFrame
+      assert(readDf.collect() sameElements writeDf.collect())
```
gengliangwang: Could you change `sameElements` to `. sameElements(writeDf.collect())` ?

gengliangwang:
Overall LGTM except one comment.

src/main/scala/com/databricks/spark/avro/AvroOutputWriter.scala
```diff
@@ -104,7 +104,13 @@ private[avro] class AvroOutputWriter(
       case DateType => (item: Any) =>
         if (item == null) null else item.asInstanceOf[Date].getTime
       case ArrayType(elementType, _) =>
-        val elementConverter = createConverterToAvro(elementType, structName, recordNamespace)
+        val elementConverter = createConverterToAvro(
+          elementType,
+          structName,
+          elementType match {
```
gengliangwang: On second thought, I think we can create a function for these four lines of code. 
A function accept elementType, recordNamespace, structName and returns the record name space.
So that we won't have duplicated code.
I am making release today, thanks for the work in advance!

gsolasab: This match is also on SchemaConverters, should we create a general function for both classes then? In wich class do you think we can put it?

gengliangwang: A general function In `SchemaConverters`, I think.

src/main/scala/com/databricks/spark/avro/AvroOutputWriter.scala
```diff
@@ -104,7 +104,11 @@ private[avro] class AvroOutputWriter(
       case DateType => (item: Any) =>
         if (item == null) null else item.asInstanceOf[Date].getTime
       case ArrayType(elementType, _) =>
-        val elementConverter = createConverterToAvro(elementType, structName, recordNamespace)
+        val elementConverter = createConverterToAvro(
+          elementType,
+          structName,
+          SchemaConverters.getNewRecordNamespace(elementType, recordNamespace, structName)
+          )
```
gengliangwang: put this `)` to the above line
https://docs.scala-lang.org/style/indentation.html#methods-with-numerous-arguments
Thanks!

gengliangwang:
Thanks, merge to master.

Issue #229, opened 2017-05-09
Read/write timestamps


sbebo:
Reading/writing dataframes to/from avro files with timestamps with spark gives an inconsistent behavior. The point being that the conversion between timestamps and integers has a different logic in spark vs. what happens working with avro: spark-avro converts timestamps to long int in _milliseconds_; however, casting them back to "timestamp" in Spark gives wildly wrong dates, as Spark expects timestamps in _seconds_ (dividing by 1000 after you read the data is an obvious workaraound), I believe.

Example:
```
scala> val df = Seq("2017-05-05", "2016-04-02").toDF.select($"value".cast("timestamp").alias("time"))

scala> df.printSchema()
root
|-- ts: timestamp (nullable = true)

scala> df.show(truncate=false)
+---------------------+
|ts                   |
+---------------------+
|2017-05-05 00:00:00.0|
|2016-04-02 00:00:00.0|
+---------------------+

scala> df.avro.write("test.avro")

scala> spark.read.avro("test.avro").show()
+-------------+
|           ts|
+-------------+
|1493935200000|
|1459548000000|
+-------------+

scala> spark.read.avro("test.avro").select($"ts".cast("timestamp")).show(truncate=false)
+----------------------+
|                    ts|
+----------------------+
|49310-12-03 17:00:00.0|
|48221-03-27 18:00:00.0|
+----------------------+

scala> spark.read.avro("test.avro").select(($"ts"/1000).cast("timestamp")).show(truncate=false)
+------------------------------+
|CAST((ts / 1000) AS TIMESTAMP)|
+------------------------------+
|2017-05-05 00:00:00.0         |
|2016-04-02 00:00:00.0         |
+------------------------------+
```

I'm wondering if there's any better way to handle this behavior than having to manually divide the timestamp fields by 1000 when you read from avro.

nmaquet:
We're being bitten by this bug, too.

ezhaar:
I agree. should be a better way to handle this. It is very easy to make mistakes with the current handling of timestamps.

tamizhgeek:
Any idea when this bug fix is getting released?

Srinathc:
Facing this issues using spark-avro 4.0.0. Please let use now the release date.

rondefreitas:
Also facing this issue writing timestamp columns from a spark dataframe.

rondefreitas:
It looks like this issue was resolved already in master.  Any chance we could get a new release for accessing via maven/sbt?

PR #253, opened 2017-11-09
Add Date and Timestamp data types support when reading avro fields

viirya:
Related issue: #229 
Related Spark JIRA ticket: https://issues.apache.org/jira/browse/SPARK-22460

Seems it is somehow inconvenient when reading avro files written with DataFrame with timestamp field.

It might be easier to read such data fields if we can explicitly require this data source to interpret a field as timestamp type. 

This also add the support of date type together.

### DIFF
--- a/src/main/scala/com/databricks/spark/avro/SchemaConverters.scala
+++ b/src/main/scala/com/databricks/spark/avro/SchemaConverters.scala
@@ -16,6 +16,8 @@
 package com.databricks.spark.avro
 
 import java.nio.ByteBuffer
+import java.sql.Timestamp
+import java.sql.Date
 
 import scala.collection.JavaConverters._
 
@@ -152,6 +154,13 @@ object SchemaConverters {
         case (IntegerType, INT) | (BooleanType, BOOLEAN) | (DoubleType, DOUBLE) |
              (FloatType, FLOAT) | (LongType, LONG) =>
           identity
+        case (TimestampType, LONG) =>
+          (item: AnyRef) => {
+            new Timestamp(item.asInstanceOf[Long])
+          }
+        case (DateType, LONG) =>
+          (item: AnyRef) =>
+            new Date(item.asInstanceOf[Long])
         case (BinaryType, FIXED) =>
           (item: AnyRef) =>
             if (item == null) {
@@ -299,7 +308,7 @@ object SchemaConverters {
         case (left, right) =>
           throw new IncompatibleSchemaException(
             s"Cannot convert Avro schema to catalyst type because schema at path " +
-              s"${path.mkString(".")} is not compatible (avroType = $left, sqlType = $right). \n" +
+              s"${path.mkString(".")} is not compatible (avroType = $right, sqlType = $left). \n" +
               s"Source Avro schema: $sourceAvroSchema.\n" +
               s"Target Catalyst type: $targetSqlType")
       }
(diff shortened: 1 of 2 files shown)


src/test/scala/com/databricks/spark/avro/AvroSuite.scala
```diff
@@ -505,6 +505,27 @@ class AvroSuite extends FunSuite with BeforeAndAfterAll {
...
+      val currentDate = new Date(System.currentTimeMillis())
+      val schema = StructType(Seq(
+        StructField("_1", DateType, false), StructField("_2", TimestampType, false)))
+      val writeDs = Seq((currentDate, currentTime)).toDS
+
+      val avroDir = tempDir + "/avro"
+      writeDs.write.avro(avroDir)
+      assert(spark.read.avro(avroDir).collect().length == 1)
+
+      val readDs = spark.read.schema(schema).avro(avroDir).as[(Date, Timestamp)]
```
saniyatech: As can be seen from this test, this PR handles the case of reading Avro when Spark (SQL) schema is known. 

For the use case where schema is to be read from the Avro file(s) (the Avro schema), this test does not verify timestamp/date is being correctly deserialized.

viirya: Yeah, this is for reading Avro with explicit Catalyst schema.

For the use case the schema is not given, we should also change the way we write the Avro file(s), IIUC.

I'd plan to try to add it in another PR. If you prefer, I can add the support here too.

viirya:
To support the use case @saniyatech mentioned, timestamp should be straightforward because it is stored as long value. We can add the logical type Timestamp (millisecond precision) to the avro's long schema.

Date maybe a real problem. I don't know why spark-avro stores date as long value of the milliseconds too. Avro's logical type Date is an Avro int. I think the logical type property isn't compatible to an avro's long schema...

viirya:
Btw, seems 1.7.6, the avro version currently used, doesn't support logical types yet. Based on above reason, I think we can only correctly deserialize date/timestamp fields when a Catalyst schema is provided.

src/main/scala/com/databricks/spark/avro/SchemaConverters.scala
```diff
@@ -152,6 +154,13 @@ object SchemaConverters {
         case (IntegerType, INT) | (BooleanType, BOOLEAN) | (DoubleType, DOUBLE) |
              (FloatType, FLOAT) | (LongType, LONG) =>
           identity
+        case (TimestampType, LONG) =>
+          (item: AnyRef) => {
```
gengliangwang: I think we should remove the brace, and make it one line

viirya: Ok.

gengliangwang:
Thanks, merge to master.

viirya:
Thanks @gengliangwang @saniyatech for review.

rondefreitas:
@gengliangwang @saniyatech any word on when this will get cut to a release? It's nearly a year since this has been fixed but it's still broken in 4.0

gengliangwang:
@rdefreitas Probably there is no new release, as this repo is migrated into Spark 2.4 as built-in data source module.

rondefreitas:
@gengliangwang any idea when that release is scheduled or where I can find that?

gengliangwang:
It will be within this October, I think.  
You can try it by building the latest apache spark (master branch or branch-2.4).

PR #121, opened 2016-03-09
Support for logical datatypes like Decimal type

cpbhagtani:
Avro doesn't support very big numbers directly. It supports it through logicalTypes where you can specify value as string type but send the actual data type of the field as logicalType. Following is the example of a decimal type

{"type": "record", 
 "name": "test",
 "fields" : [
 {"name": "a","type": "string",
  "logicalType": "decimal",
  "precision": 4,
  "scale": 2
},
 {"name": "b", "type": "string"}
]}

This pull request add support for reading such datatype. For now I have added only decimal type, but we can add more logicalTypes. In dataframe it actually appears as decimal(4,2) type.

PR #257, opened 2017-11-27
Support for logical datatype like decimal type

varsha1990:
This pull request is in context to - https://github.com/databricks/spark-avro/pull/121 which was done by @cpbhagtani to support the logical types. This current PR adds this support for spark 2.x


gengliangwang:
In [Avro spec](https://avro.apache.org/docs/1.8.2/spec.html):
```
A decimal logical type annotates Avro bytes or fixed types
```
Can you handle `bytes` and `fixed` type as well?

cbia4:
Is there any movement on this? I have yet to find a way to load avro data into a data frame where data with type bytes (and logical type decimal with scale and precision) is converted back to a human readable form.

cbia4:
I updated the avro package version and rebuilt SchemaConverters.scala with something similar to what is in the PR - except like I mentioned, Avro 1.8.2 stores the decimal logical type as bytes, and you need to use Conversions.DecimalConversion from the avro API to make it happen, but it works as expected for anyone who needs to get this to work with decimal types.

peter-ungvari:
@cbia4 Where can I find your changes?
Is there another PR for bytes avro type?

I am ready to work on this in order to make spark-avro compatible with avro spec as well as Hive regarding to decimal types.

LuboVarga:
I wish this project is in maven... I am more familiar with it. But I can promise to share result of my work (decimal type for bytes and fixed types and perhaps timestamp logical type)

koertkuipers:
sbt has a funny syntax but on the console it pretty easy to use as a
developer. its compile and test cycle is much faster than maven. you will
be happy with all the time you get back not spending waiting on super slow
maven:)

PR #263, opened 2017-12-28
Upgrade SBT and corresponding plugins.

aa8y:
This PR upgrades SBT to 0.13.16. It could not be upgraded to the latest version (1.0.4 as of now) because of [this issue](https://github.com/databricks/sbt-spark-package/issues/35) with `sbt-spark-package`. `sbt-idea` was removed from the plugins because, firstly, it is very old and would not support SBT 1+ and also because IntelliJ now [supports SBT](https://blog.jetbrains.com/scala/2017/03/23/scala-plugin-for-intellij-idea-2017-1-cleaner-ui-sbt-shell-repl-worksheet-akka-support-and-more/) which should be the correct way.

### DIFF
--- a/project/build.properties
+++ b/project/build.properties
@@ -14,4 +14,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-sbt.version=0.13.13
+sbt.version=0.13.16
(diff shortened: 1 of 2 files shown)


aa8y:
@gengliangwang Do you mind taking a look at this?

gengliangwang:
Thanks, merge to master

PR #269, opened 2018-02-12
Pick first file sorted by path for schema

koertkuipers:
Picking the same file consistently for schema avoids weird bugs where the schema of an avro data source changes randomly or unexpectedly.

### DIFF
--- a/src/main/scala/com/databricks/spark/avro/DefaultSource.scala
+++ b/src/main/scala/com/databricks/spark/avro/DefaultSource.scala
@@ -61,24 +61,28 @@ private[avro] class DefaultSource extends FileFormat with DataSourceRegister {
       files: Seq[FileStatus]): Option[StructType] = {
     val conf = spark.sparkContext.hadoopConfiguration
 
-    // Schema evolution is not supported yet. Here we only pick a single random sample file to
+    // Schema evolution is not supported yet. Here we only pick the first file sorted by path to
     // figure out the schema of the whole dataset.
-    val sampleFile = if (conf.getBoolean(IgnoreFilesWithoutExtensionProperty, true)) {
-      files.find(_.getPath.getName.endsWith(".avro")).getOrElse {
-        throw new FileNotFoundException(
-          "No Avro files found. Hadoop option \"avro.mapred.ignore.inputs.without.extension\" is " +
-            "set to true. Do all input files have \".avro\" extension?"
-        )
-      }
+    def sampleFilePath = if (conf.getBoolean(IgnoreFilesWithoutExtensionProperty, true)) {
+      files.iterator.map(_.getPath).filter(_.getName.endsWith(".avro"))
+        .reduceOption{ (p1, p2) => if (p1.compareTo(p2) <= 0) p1 else p2 }
+        .getOrElse {
+          throw new FileNotFoundException(
+            "No Avro files found. Hadoop option \"avro.mapred.ignore.inputs.without.extension\" " +
+              "is set to true. Do all input files have \".avro\" extension?"
+          )
+        }
     } else {
-      files.headOption.getOrElse {
-        throw new FileNotFoundException("No Avro files found.")
-      }
+      files.iterator.map(_.getPath)
+        .reduceOption{ (p1, p2) => if (p1.compareTo(p2) <= 0) p1 else p2 }
+        .getOrElse{
+          throw new FileNotFoundException("No Avro files found.")
+        }
     }
 
     // User can specify an optional avro json schema.
     val avroSchema = options.get(AvroSchema).map(new Schema.Parser().parse).getOrElse {
-      val in = new FsInput(sampleFile.getPath, conf)
+      val in = new FsInput(sampleFilePath, conf)
       try {
         val reader = DataFileReader.openReader(in, new GenericDatumReader[GenericRecord]())
         try {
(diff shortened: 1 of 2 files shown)


src/main/scala/com/databricks/spark/avro/DefaultSource.scala
```diff
@@ -61,24 +61,28 @@ private[avro] class DefaultSource extends FileFormat with DataSourceRegister {
...
     // figure out the schema of the whole dataset.
-    val sampleFile = if (conf.getBoolean(IgnoreFilesWithoutExtensionProperty, true)) {
-      files.find(_.getPath.getName.endsWith(".avro")).getOrElse {
-        throw new FileNotFoundException(
-          "No Avro files found. Hadoop option \"avro.mapred.ignore.inputs.without.extension\" is " +
-            "set to true. Do all input files have \".avro\" extension?"
-        )
-      }
+    def sampleFilePath = if (conf.getBoolean(IgnoreFilesWithoutExtensionProperty, true)) {
+      files.iterator.map(_.getPath).filter(_.getName.endsWith(".avro"))
```
gengliangwang: files.map(_.getPath).sortBy(_.getName)

src/main/scala/com/databricks/spark/avro/DefaultSource.scala
```diff
@@ -61,24 +61,28 @@ private[avro] class DefaultSource extends FileFormat with DataSourceRegister {
...
+          throw new FileNotFoundException(
+            "No Avro files found. Hadoop option \"avro.mapred.ignore.inputs.without.extension\" " +
+              "is set to true. Do all input files have \".avro\" extension?"
+          )
+        }
     } else {
-      files.headOption.getOrElse {
-        throw new FileNotFoundException("No Avro files found.")
-      }
+      files.iterator.map(_.getPath)
```
gengliangwang: ditto

src/main/scala/com/databricks/spark/avro/DefaultSource.scala
koertkuipers: it has same result right?

files can be a very large sequence. the iterator approach avoids creating 2 copies of that sequence. also it is not necessary to do a full sort just to get the first sorted element.

 are you saying its not worth the optimization?

src/test/scala/com/databricks/spark/avro/AvroSuite.scala
```diff
@@ -803,4 +803,15 @@ class AvroSuite extends FunSuite with BeforeAndAfterAll {
...
     }
   }
+
+  test("writing avro partitions with different schemas and reading back out with a single predictable schema") {
+    TestUtils.withTempDir { tempDir =>
+      val df1 = spark.createDataFrame(Seq(("a", 1), ("b", 2)))
+      df1.write.avro(s"$tempDir/different_schemas/z=1")
+      val df2 = spark.createDataFrame(Seq(Tuple1("a"), Tuple1("b")))
+      df2.write.avro(s"$tempDir/different_schemas/z=2")
+      val df3 = spark.read.avro(s"$tempDir/different_schemas")
```
gengliangwang: maybe add a loop for the reading? I am sure if the order will be different every time

src/main/scala/com/databricks/spark/avro/DefaultSource.scala
gengliangwang: You are right for not sorting all the file names.
But I don't think we need to convert it to an `iterator`. 
Maybe we can try to make it more shorter like `files.map(_.getPath).minBy(_.getName)`  ?

koertkuipers: iterator is lightweight and avoids materialization

minBy(_.getName) wouldnt work because we want to sort by the path, not just the filename (e.g. `/some/path/x=1/part-0000.avro` comes before `/some/path/x=2/part-0000.avro`)

minBy(_.toString) might work but i don't feel too certain about it. rather use Comparable to do the right thing. unfortunately Path is just Comparable, not Comparable[Path], so scala doesn't understand how to use it, which is why i resorted to using compareTo directly.

cwlaird3:
Have you considered using the schema from the newest data file to get the most up to date version of the schema?  Or perhaps a configuration option to do that?  Seems like most would update their schemas in a backwards compatible way and  using the most recent schema would expose newer fields in the schema.

koertkuipers:
t
​hat is not a bad idea. a switch seems reasonable. i would suggest to do
this in a separate branch​

gengliangwang:
@cwlaird3 good idea
@koertkuipers how about by default use the latest AVRO file's schema?

gengliangwang:
@koertkuipers @cwlaird3 I checked with @liancheng , which is PMC member and one of the original author of Data source project. 
He doesn't think we should make such assumption. If the schema is different among files,  users are supposed to specify the schema:
https://spark.apache.org/docs/latest/sql-programming-guide.html#programmatically-specifying-the-schema

This PR changes the behavior and is possible to cause regression to other users.

koertkuipers:
currently it uses a random file to pick schema. what would a users for which you break things by going from a random file to the last file?

cwlaird3:
I agree with @koertkuipers .. but if there's still a concern adding a configuration option to change the behavior could address that.

koertkuipers:
spark-avro already provides a mechanism for the user to provide a schema with the `avroSchema` key in options

the thing that is currently missing is merging of schemas across all files

cwlaird3:
By configuration I meant a flag to enable the behavior you've implemented here - not to provide a schema.

koertkuipers:
oh a flag to go from random schema to non-random schema?
if someone can come up with a user for which this pullreq breaks their
usage i am up for that, otherwise no :)

PR #271, opened 2018-03-04
 TimeUnit conversion for BigQuery compatibility

devlucasc:
These changes allow to set option that make conversion of timestamp in long format using a specified precision, for example, microseconds.
The setting of precision option can make avro compatible with BigQuery format, that reads timestamp in microseconds.


gengliangwang:
Hi @devlucasc ,
In my opinion:
1. If we should add this option, we need to add it for all the data sources, instead of just for AVRO.
2. We can use `.map` or SQL statement to change the output timestamp unit, which is simple and straight forward to me.

devlucasc:
Hi @gengliangwang 
1) The problem only occur in avro data sources
2) Map is not good when has a complex structure as arrays, structs or array of structs that contains a timestamp field. In this case is needed to make map, flatmap and explode to treat.

Kind regards.

PR #286, opened 2018-07-14
 Do not ignore avro files without extensions by default

MaxGekk:
## What changes were proposed in this pull request?

In the PR, I propose to change the default behaviour of AVRO datasource which currently ignores files without `.avro` extension in read by default. This PR sets the default value for `avro.mapred.ignore.inputs.without.extension` to `false` in the case if the parameter is not set by an user.

## How was this patch tested?

Added a test file without extension in AVRO format, and new test for reading the file with and wihout specified schema.


gengliangwang:
Left some comments in https://github.com/apache/spark/pull/21769 .  
Turn out there is bug on reading this option.

MaxGekk:
@gengliangwang I ported the commit https://github.com/apache/spark/commit/ba437fc5c73b95ee4c59327abf3161c58f64cb12 . Please, take a look at the PR again.

gengliangwang:
Thanks, merge to master.

